PR #2, opened 2019-07-04
search rotated sorted array and unknown length array completed -please review

PremalSamale:


jaspindersingh83:
Great job Premal
1) int median =left+(right-left)/2; This might not be median in our case. Try recursion here as well
2) Good job with problem 2

PR #3, opened 2019-07-04
Please review Day 3 : Solved rotated sorted array problem using binary search

MahiteeS:


jaspindersingh83:
1) This is a different but a good way of doing problem 1. We are doing binary search twice though
2) Problem 2: Good. Add exception handling

PR #4, opened 2019-07-05
Updated Day3 Solutions after the lecture 

gauravj09:


jaspindersingh83:
Good job with problem 1 Gaurav 
1) int middle = (left+right) / 2; can be rewritten like we discussed in class. Try recursive solution as well
2) Complete the problem 2

PR #5, opened 2019-07-05
Completed Day 3 - Kindly Review

Hariae:


jaspindersingh83:
1) Sorted pivot is also a good solution
2) Add exception handling in problem #2

PR #13, opened 2019-07-31
Binary Search. Please review

priyansh22:
There is some issue with the code. Works in some cases; doesn't work in some cases.

Like in the example : 4,5,6,7,0,1,2. 
If the element searched is 4. It shows found(Same with 4,5,6,7)
If the element is searched is 0,1,2. It shows not found.

But in the example : 7,0,1,2 (If you change the input)
If you search any element from the above, it shows as found.

The "not found" case is also not working.

Please review

### DIFF
--- a/binarysearch.cpp
+++ b/binarysearch.cpp
@@ -1,0 +1,38 @@
+#include<iostream>
+using namespace std;
+
+class BinarySearch
+{
+public:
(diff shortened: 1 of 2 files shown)


priyansh22:
Ok yes I have understood the mistake now. Will correct the code

PR #22, opened 2019-08-21
Binary Search-1 done

swaps95shah:


problem2.py
```diff
@@ -0,0 +1,29 @@
...
+time: O(logn)
+space: O(1)
+Leet : Problem not from leetcode
+Problems faced: No problems.
+"""
+def infiniteBinary(nums,target):
+    slow = 0
+    fast = 1
+    
+    while nums[fast] < target:
```
kaurramanpreet: what if we reach out of range of the given array because here infinite also means a particular value right?

swaps95shah: You are right. My program doesn't handle that. I will review the problem and submit again soon. Good catch!

swaps95shah:
Corrected problem2.py
Added linear search where an exception is thrown

PR #29, opened 2019-10-09
initial check in - one of the homework

vramanathan:
problem 5. others in progress

### DIFF
--- a/RotatedSortedArray.java
+++ b/RotatedSortedArray.java
@@ -1,0 +1,100 @@
+package test11;
+
+public class RotatedSortedArray {
+
+	// Returns index of x if it is present in arr[l.. r], else return -1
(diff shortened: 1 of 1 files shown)


RotatedSortedArray.java
```diff
@@ -0,0 +1,100 @@
...
+    				// explore left half
+    				r = mid - 1;
+    			}
+    		} else if (arr[l] > arr[mid]) {
+    			// left side is unsorted, search here
+    			r = mid - 1;
+    		} else if (arr[mid] > arr[r]) {
+    			// right side is unsorted, search here
+    			l = mid + 1;
+    		}
```
sehaj-grover: these cases are unnecessary if we have already checked the sorted cases

PR #42, opened 2019-11-20
search element in pivoted array

AvinavTyagi:


AvinavTyagi:
I completed both problems and updated the PR

PR #47, opened 2019-11-20
Pushing Design 3 Changes - Prateek

dasgprat:
Please review the PR and also let me know in case of any issue or concerns.


dasgprat:
Thanks for reviewing.
I already checked in optimal codes in the same PR for Rotated matrix after class.

PR #64, opened 2019-12-25
Question 5 and 6 using Binary Search(Please Review)

ShreyasP7495:


ques6.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+int findpos(int a[],int key)
+{
+    int l=0,h=1;
+    int val=a[h];
+    while(val<key)
+    {
+        l=h;
+        h=2*h;
```
kaurramanpreet: how will you handle exceptions in this case?

PR #66, opened 2019-12-26
Binary Search-1: Please Review

BhushanVasisht:


Problem2.java
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	private static int modBinSearch(int[] nums, int target)
+	{
+		int l=0, h=1;
+
+		while(h < nums.length && target >= nums[h])
+		{
+			l = h;
+			h = 2*h;
```
kaurramanpreet: how will you handle the case when h is less than nums.length but when this condition will be satisfied we will have h as 2*h but in some cases 2*h will be out of range

PR #67, opened 2019-12-26
Binary Search in Rotated array

nehabindle:
Please review

### DIFF
--- a/RotatedArray.java
+++ b/RotatedArray.java
@@ -1,0 +1,74 @@
+package RotatedArray;
+
+class Solution {
+    public static int search(int[] nums,int target) {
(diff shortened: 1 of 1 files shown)


RotatedArray.java
```diff
@@ -0,0 +1,74 @@
+package RotatedArray;
+
+class Solution {
+    public static int search(int[] nums,int target) {
+    	int left = 0;
+    	int right = nums.length-1;
+    	
+    	while(left<= right && left<nums.length &&  right>=0)
```
kaurramanpreet: why do you need these 3 conditions
when only left<=right can work

PR #72, opened 2019-12-27
Completed Binary_Search-1. Please Review. 

habildamania:
@rajatkumar01


habildamania:
Figured out what I was doing wrong in the 2nd Problem. My program was correct because my code logic was assuming it to be an infinite array. So when my target element was near the end of the sample array that I provided, the high pointer was being set to an index out of the array. So it was giving an error. But in theory if we assume it to be an infinite array, it should not give the error.

PR #82, opened 2020-01-03
Completed Search in a Rotated Sorted Array and Search in an Infinite Array - Please review

ashu20red:


ashu20red:
Hi,
Yes. It results in integer overflow problem. So using mid=(low+high)/2 will
increase the range by shifting towards the negative integer -214,.... which
can result in error.

Please correct me if I am wrong.

super30admin:
Yup you are right

PR #86, opened 2020-02-05
Completed BinarySearch-1, please review

sowmyadvn:


InfiniteSortedArraySearch.java
```diff
@@ -0,0 +1,42 @@
+// Time Complexity : O(log N) where N is size of the array
+// Space Complexity : O(1) auxillary space as we haven't used additional space for the logic
+// Did this code successfully run on Leetcode : Ran it on leetcode playground
+// Any problem you faced while coding this : Because I took a finite array test-case, if I was searching for element in the end of array
+//, threw ArrayIndexOutOfBoundsException as high*2 index is not available. (2340 for example)
+
```
super30admin: That exception is because of the test cases we take. In actual there will be an infinite array

sowmyadvn: Got it, thank you.

PR #98, opened 2020-02-07
Completed Binary Search

RajiBabuMSSE:

I have some questions for InfiniteSorted Array. I have commented those. Kindly check it.

### DIFF
--- a/SearchInfiniteSortedArray
+++ b/SearchInfiniteSortedArray
@@ -1,0 +1,71 @@
+// Time Complexity  - O(log n)
+// Space Complexity - 0(1) - Constant space
+// Did this code successfully run on Leetcode :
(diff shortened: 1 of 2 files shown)


SearchInfiniteSortedArray
```diff
@@ -0,0 +1,71 @@
+// Time Complexity  - O(log n)
+// Space Complexity - 0(1) - Constant space
+// Did this code successfully run on Leetcode :
+// Any problem you faced while coding this : Input - {3, 5, 7, 9, 10, 90, 100, 130, 140, 160, 170}
+// Yes, when I passed 160 and 170, I am getting ArrayIndexOutofBoundsException
+
```
super30admin: Yes that would happen. As the example used is a finite array. However in practice it will be an infinite array and we won't go out of bound ever

PR #105, opened 2020-02-10
Done with problems. Please review and provide feedback

singhperryo1:


Problem2.java
```diff
@@ -0,0 +1,45 @@
...
+		else if (a[m] < target) l = mid + 1 ; 
+	}
+
+	return -1 ; 
+
+	}
+}
+
+
+
```
super30admin: But where are we applying both these functions. The main function is missing here

PR #111, opened 2020-03-12
Please review

megrao:
Not sure of the second solution. Logic might be correct. There is something wrong in the code.


devashishnyati:
Good job Megha!
Yes, the time complexities for both the problems are O(logn). At every step, you are removing half of the elements. This is not constant time.
The space complexity is O(1) because you are not using any extra space.

PR #122, opened 2020-03-15
Binary Search - I

nilpatel111296:


devashishnyati:
Good job Nil!
Please add the time and space complexities along with whether the code ran successfully on leetcode and challenges faced at the top.

PR #127, opened 2020-03-30
BS-1 is complete. Please review and share feedback.

techielove92:
Some doubts in infinite array problem.

### DIFF
--- a/Problem2.py
+++ b/Problem2.py
@@ -1,0 +1,61 @@
+"""
+#Given a sorted array of unknown length and a number to search for, return the index of the number in the array.
+# Accessing an element out of bounds throws exception. If the number occurs multiple times, return the index of any occurrence.
+# If it isn’t present, return -1.
+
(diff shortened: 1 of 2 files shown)


Problem2.py
```diff
@@ -0,0 +1,61 @@
...
+    def Search_Infinite(self,nums,target):
+        low=0   # intialise low as 0
+        high=2   # high as 2
+        try:     #The try block lets you test a block of code for errors. The except block lets you handle the error.
+            while nums[high]<target:   # if target is less than nums[high] we got to except part
+                low=high
+                high=high*2
+        except Exception as e:
+            try:        #
+                high=low+1
```
devashishnyati: This will make it linear.

devashishnyati: You don't have to worry about having the number in the last bucket because it's an infinite array.
Still you can refer https://github.com/devashishnyati/Binary-Search-1/blob/master/unkownlength.py

devashishnyati:
Good job Lovenish.

PR #135, opened 2020-04-23
Binary Search -1 Completed -- Please Review 

jeelchristian:


Problem#6.txt
```diff
@@ -0,0 +1,65 @@
...
+        
+        // while the val if lesser than the target we keep on moving the start and end pointers ahead
+        
+        while( val < target){
+            
+            start = end;   // new start is equal to end
+            
+            // Here we are going to assign end as 2 times of end or array length which ever is smaller
+            
+            if( 2*end < arr.length -1)
```
super30admin, line 31: We don't have access to arr.length

Problem#6.txt
```diff
@@ -0,0 +1,65 @@
...
+        return Search( arr,  start, end,  target);
+     }
+     
+     public static int Search(int[] arr, int start, int end, int target){
+         
+         // itterate till start is less than end
+          while(start < end){
+              
+              // here mid should be between start and end so start is added to (start+end)/2
+              int mid  = start + ((start + end) / 2);
```
super30admin, line 51: Review this step again. This should not be working

super30admin:
Kindly review the inline comments.

Need more focused coding.

Problem#6.txt
jeelchristian: But for the temporary purpose, it was a hypothetical situation so while testing my code I got an array index out of bounds error so I added this line.

Problem#6.txt
jeelchristian: In Code, it was working, but after class I got the idea that it should be start + (end-start)/2;

PR #137, opened 2020-04-23
Completed Binary-Search-1 - Please Review

karthikmk:


karthikmk:
Sorry about that. Please check the updated pull request.

rotatedSortedArray.cc
```diff
@@ -1,26 +1,30 @@
+// Time Complexity : O(log n)
+// Space Complexity : O(log n)
```
super30admin, line 2: Better to go with iterative binary search. We can achieve O(1) space

super30admin:
Rather than calling Binary search recursive function again and again...Call recursive function only once at the end of function with updated low and high pointer values 

Please review inline comment as well

PR #138, opened 2020-04-23
Jay_Nirmal_PR

jaynirmal15:


FindElementInInfiniteArray.java
```diff
@@ -0,0 +1,64 @@
...
+	       
+	    }
+	   
+	      static int search(int num[], int low, int high, int target)
+	    {
+	         
+	        if (low > high) 
+	            return -1;
+	       
+	        int mid = (low+high)/2;
```
iivaishnavii, line 38: In order to not face Integer overflow error, we generally calculate mid as

 mid = low + (high-low)/2

FindElementInInfiniteArray.java
```diff
@@ -0,0 +1,64 @@
...
+		// TODO Auto-generated method stub
+		public int search(int[] nums, int target) {
+			
+			
+			int length = 1;
+			
+			//calculate the high end for the array search since we don't know the length
+			try { while(target > nums[length])
+		     {
+		    	 length = length+1;
```
iivaishnavii, line 19: You can increment it by steps of two,as an optimization.

FindElementInInfiniteArray.java
jaynirmal15: Yeah, I got. I learned about it yesterday during the lecture. I will update it.

iivaishnavii:
Great job,Nirmal. Its impressive to see how you have handled exceptions as well.

Going forward, do mention time and space complexities for all solutions.

jaynirmal15:
Thank you for the comment Vaishnavi,  I will keep that in mind for future solutions.

PR #139, opened 2020-04-23
Completed Design Please review

JacobStephen9999:


iivaishnavii:
Good job, Jacob. Great Attempt!All your conditions were correct for searching an element in rotated sorted array.

For infinite array, if the element is not present in the array, then you will have an infinite loop. Always keep the edge cases in mind as well. Don't forget to submit the revised solution after class.

PR #156, opened 2020-05-17
Binary-Search-1

adityad30:


adityad30:
For the problem search target in unknown length(Problem 6), how can we handle the case when value of high goes beyond the length of an array ? The code fails for the values towards the end of an array. 
pointers are updated as : 
slow = high 
high = high * 2

sowmyadvn:
> For the problem search target in unknown length(Problem 6), how can we handle the case when value of high goes beyond the length of an array ? The code fails for the values towards the end of an array.
> pointers are updated as :
> slow = high
> high = high * 2

@adityad30  Possible test-cases we can check our code on are finite, so it might lead to ArrayOutOfBoundsException but the question is regarding infinite-sized array . So, it is fine if the code fails for finite size arrays

sowmyadvn:
Good job Aditya! All the solutions are correct! 
Please review inline comment regarding failed test-case

sowmyadvn:
Good job Aditya! All the solutions are correct! 
Please review inline comment regarding failed test-case

adityad30:
Thank you!

PR #158, opened 2020-05-26
Completed : Please Review

ashiranka:


InBinSearch.java
```diff
@@ -0,0 +1,53 @@
+// Time Complexity : O(sqrt(n)) n is number of elements in array, sqrt(n) : to find r length
```
chiragshahi, line 1: log(n) where n is the position of the target number

chiragshahi:
Great job with code and comments! See inline comment

PR #161, opened 2020-05-27
Completed binary search 1 - Took external help ( Google)

GahanJagadeesh:


InfiniteArray.java
```diff
@@ -0,0 +1,50 @@
+// Time Complexity : O(log(Position))
```
Lohith-Aswathappa, line 1: It'll be better if you could generalize this or explain the variable used

RotatedArray.java
```diff
@@ -0,0 +1,38 @@
...
+
+// Your code here along with comments explaining your approach
+
+class Solution {
+    public int search(int[] nums, int target) {
+        int left=0;
+        int right=nums.length-1;
+        
+        while(left<=right){
+            int mid = (left+right)/2;
```
Lohith-Aswathappa, line 15: Sometimes this might lead to integer overflow, practice using (left + (right-left)/2) option

Lohith-Aswathappa:
Good work, Gahan!
Review inline comments.

PR #162, opened 2020-05-27
Commit1

PadmaPriyaCheruvu:
1. UnknownArray - Find target in array of unknown length. In this code, accessing an element out of bounds throws exception.

2. UnknownArrayWithArrayReader - This code uses ArrayReader as a supporting class as found on Leetcode. Here, accessing an element out of bounds does not throw an exception.

### DIFF
--- a/SearchInRotated.java
+++ b/SearchInRotated.java
@@ -1,0 +1,33 @@
+// Time Complexity : O(log n)
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+
+class SearchInRotated {
+    public int search(int[] nums,int target){
+        int s=0,e = nums.length-1;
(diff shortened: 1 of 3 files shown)


SearchInRotated.java
```diff
@@ -0,0 +1,33 @@
+// Time Complexity : O(log n)
```
chiragshahi, line 1: define n

UnknownArray.java
```diff
@@ -0,0 +1,54 @@
+// Time Complexity : O(log t) where t is index of target
```
chiragshahi, line 1: Great job figuring this out

chiragshahi:
Excellent clean code with comments!

SearchInRotated.java
PadmaPriyaCheruvu: n is the input array size

PR #163, opened 2020-05-27
Search in rotated sorted Array

mithi21:
Use direct Binary search approach
Space complexity : O(1)
time complexity : O(N)


Search_infinite_array.java
```diff
@@ -0,0 +1,35 @@
+/*
+    Time Complexity :  O(log(N)) N =Size of input
```
chiragshahi, line 2: n is the position of the target number

chiragshahi:
Good job! interesting to see submissions in different languages

Search_infinite_array.java
mithi21: changed please check updated one

PR #166, opened 2020-05-27
Completed Binary-Search-1. Kindly review and share you feedback.

Sumesh-H:


InfiniteBinarySeach.java
```diff
@@ -0,0 +1,54 @@
+// Time Complexity : O(log n)
```
chiragshahi, line 1: define n

InfiniteBinarySeach.java
```diff
@@ -0,0 +1,54 @@
...
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : When I try to find an element which
+   is not present in the array after the 1st index, StackOverflowError is shown by the compiler.
+
+// Your code here along with comments explaining your approach
+
+class BinarySearch {
+    //Reglur binary search
+    static int binarySearch(int arr[], int l, int r, int x){
+        if (r>=1){
```
chiragshahi, line 12: why is r >= 1

chiragshahi, line 12: why is r >= 1?

RotatedBinSearch.java
```diff
@@ -0,0 +1,47 @@
...
+                return midpoint;
+            }else if (nums[midpoint] < target){
+                l = midpoint + 1;
+            }else {
+                r = midpoint - 1;
+            }
+        }
+        return -1;
+    }
+}
```
chiragshahi, line 47: See this code which is easier to explain in an interview
public int search(int[] arr, int target) {
        int low = 0, high = arr.length - 1;
        while (low <= high) {
            if (arr[mid] == target) return mid;
            // if left half sorted
            if (arr[low] <= arr[mid]) {
                if (target < arr[mid] && target >= arr[low]) high = mid-1;
                else low = mid+1;
            } else {
                if (target > arr[mid] && target <= arr[high]) low = mid+1;
                else high = mid-1;
            }
        }
        return -1;

chiragshahi:
Good job! Kindly check the inline comments

InfiniteBinarySeach.java
Sumesh-H: I will correct it. It was a mistake.

RotatedBinSearch.java
Sumesh-H: Yeah, I understood the approach in the lecture. I will keep this in mind

InfiniteBinarySeach.java
Sumesh-H: n is the index of the target in an array

PR #173, opened 2020-05-27
Please review - First problem completed

Abhishek1004:


infarraysearch.py
```diff
@@ -0,0 +1,21 @@
+def interval_det(arr, lower_bound, upper_bound, target):
+	if target<upper_bound and target>lower_bound:
+		return binary_search(arr, lower_bound, upper_bound, target)
+	else:
+		return interval_det(arr, lower_bound+200, upper_bound+200, target)
```
AbhiniveshP, line 5: This would change the span or interval of search linearly and not logarithmically. For the solution to be in O(logn), try to change the span by a factor of 2.

AbhiniveshP:
Kindly include a 3-step description, time, and space complexities, along with writing some extra comments wherever necessary.

infarraysearch.py
Abhishek1004: Oh yes. It was mentioned in class. I did this before that. Will update this

PR #175, opened 2020-05-27
Binary Search Exercise - Please review it

nareshkumar-ss:


AbhiniveshP:
Kindly include a 3-step solution that explains your approach, along with Time and Space complexities; and whether the code ran all test cases or not. 
Also including comments wherever they have will help you in case you need to revisit this question in the future.
Along with these, also try iterative approaches for both these questions.

PR #177, opened 2020-05-27
Finished sprint Binary Search -Please Rivew

osamamohamedsoliman:


SearchRotatedSortedArray.py
```diff
@@ -0,0 +1,51 @@
...
+def search(nums, target):
+    """
+    :type nums: List[int]
+    :type target: int
+    :rtype: int
+    """
+    start = 0
+    end = len(nums)-1
+    while start <=end:
+        mid = (start+end)//2
```
Lohith-Aswathappa, line 18: practice using (start + (end -start)/2). it'll help you avoid integer overflow error.

Lohith-Aswathappa:
Good job, Osama!

PR #179, opened 2020-05-27
COmpleted - Please review

shivanireddy:


SearchSortedRotatedArray.java
```diff
@@ -0,0 +1,32 @@
+// Time Complexity : O(log N) average
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Not sure about the space complexity. I'm using recursion so I think that the space complexity isn't O(1)
```
Lohith-Aswathappa, line 4: if you don't consider the recursive stack, it'll be O(1)
if you consider, you'll be storing the return address to atmost logN times.

Lohith-Aswathappa:
Good work, Shivani!
Try to add comments wherever necessary, it'll help you review later.

SearchSortedRotatedArray.java
shivanireddy: Thank you. Then it'll be O(log N) in this case because I considered recursion stack.

PR #181, opened 2020-05-27
Created Binary-Search-1.py

Shashank001122:


Lohith-Aswathappa:
Good work, Shashank!
Try the infinite search problem and also add time and space complexities to all the problems.

PR #199, opened 2020-05-29
Ktc binary search1 complete

Keerti995:


src/SearchRotatedArray.java
```diff
@@ -0,0 +1,62 @@
+// Time Complexity : O(log n)
```
chiragshahi, line 1: define n

src/SortedUnknownLengthArray.java
```diff
@@ -0,0 +1,72 @@
+// Time Complexity : O(log T) where T is the target value
```
chiragshahi, line 1: T is the index of the target value

chiragshahi:
Excellent work Keerti! See inline comments

PR #212, opened 2020-07-09
Binary Search-1 : Please Review

akashkshekarappa:


src/SearchInaRotatedArray.java
```diff
@@ -0,0 +1,102 @@
...
+    public int search(int[] nums, int target) {
+        
+        if(nums == null || nums.length == 0)
+            return -1;
+        
+        return binarySearch(nums, 0, nums.length-1, target);
+    }
+    
+    public int binarySearch(int[] nums, int low, int high, int target){
+         while(low <= high){
```
samarthjj, line 68: This will work with **while** too but, this should be "if low <= high".

src/SearchInaRotatedArray.java
```diff
@@ -0,0 +1,102 @@
...
+			System.out.println("Not found");
+		else
+			System.out.println("Found at index " +index);
+	}
+}
+
+/*-----------------------------------------Recursive Approach-------------------------------------------------------*/
+
+/*
+ class Solution {
```
samarthjj, line 58: Please state the Time & Space complexity for every approach you mention.

samarthjj:
Good work Akash, please review the inline comments.

PR #215, opened 2020-07-09
Adding solution for Binary Search-1

goelsarthak:
Please find below the implementation of below problems:

1. Find element in rotated sorted array
2. Find element in sorted array whose length is unknown.

Note: Time Complexity of each problem is stated at top of each program.


Sumesh-H:
Good work, Sarthak! Try explaining your approach in 3-4 bullet points before each solution.

PR #219, opened 2020-07-09
Completed BinarySearch1 - search in rotated sorted array and unknown-sized array

alekhya-12:
- Implemented the feedback received from the previous PRs by maintaining two separate files.
- Also, wrote 3-point approaches and Time and Space Complexities as well as I could.
- Please review. Thanks!

### DIFF
--- a/BinarySearchUnknownSize.java
+++ b/BinarySearchUnknownSize.java
@@ -1,0 +1,53 @@
+// Time Complexity : O(log n) where n is the number of elements in the array, but here we don't know what n is.
+// Space Complexity :  Space complexity is the size of the array itself.
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this :
+
+
+// Your code here along with comments explaining your approach
+// Approach :
+// 1. Make use of two pointers. Slow pointer and fast pointer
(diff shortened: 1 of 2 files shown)


BinarySearchUnknownSize.java
```diff
@@ -0,0 +1,53 @@
+// Time Complexity : O(log n) where n is the number of elements in the array, but here we don't know what n is.
+// Space Complexity :  Space complexity is the size of the array itself.
```
samarthjj, line 2: The space complexity is O(1), since no additional space apart from the pointers used.
Input and Output Data Structures do not qualify as "Additional Space".

RotatedSortedArray.java
```diff
@@ -0,0 +1,115 @@
+// Time Complexity : O(log n), where n is the size of the array 
+// Space Complexity : O(1) // since it's a finite sized array
```
samarthjj, line 2: The space complexity should be O(1). 
Your reasoning here is incorrect, it is because no additional space apart from the pointers was used.

samarthjj:
Good work Shilpi, please review the inline comment.
Good work on including both the recursive and the iterative solution.

samarthjj:
Good job Alekhya, Please review the inline comment.
Please mention Time & Space Complexity for every approach mentioned as well.

PR #222, opened 2020-07-10
problem 1 is completed. please review. 

satishsolanki1990:
Hi, Sorry, I couldn't update the sample.java file. I will update this weekend. It was a busy week. thanks.


samarthjj:
Good job Satish, please try the other problem as well. 
You don't have to update the sample.java file, use that template (in the same file as the solution) to mention time and space complexity.

samarthjj:
Good job Satish, please try the other problem as well.
You don't have to update the sample.java file, use that template (in the same file as the solution) to mention time and space complexity.

PR #224, opened 2020-07-10
Completed Binary-Search-1. Please review!

arrpitsharrma:


search-sorted-array-unknown-size.py
```diff
@@ -1,3 +1,13 @@
+# Time Complexity : O(n), worst case, may be. Not sure, as the array is of unknown size.
```
Sumesh-H, line 1: Even if it is unknown TC will be O(log N) where n is the number of elements in the array.

Sumesh-H:
Good work, Arrpit! Check inline comments.

PR #231, opened 2020-07-10
Binary Search 1 completed

Balaviknesh:
I dont have leet code premium account for problem 2. Please let me know how I can get the problem. thanks.

### DIFF
--- a/RotatedSortedArray.py
+++ b/RotatedSortedArray.py
@@ -8,24 +8,25 @@ class Solution:
 
     def search(self, nums: List[int], target: int) -> int:
 
-        #Intializing the left and right values which is start and end, In python we can give this by 0(start) and len(nums)-1 (end)
+        # Initializing the left and right values which is start and end, In python we can give this by 0(start) and
+        # len(nums)-1 (end)
         left = 0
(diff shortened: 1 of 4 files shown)


RotatedSortedArray.py
```diff
@@ -0,0 +1,99 @@
...
+
+                    right = mid - 1
+
+        return -1
+
+
+# Recursive
+
+# Time Complexity : O(log(n))
+# Space Complexity : O(n) - I have no clear explanation about this. This is just a random guess.please give me the explanation for this. Thanks.
```
Sumesh-H, line 53: When we use a recursive approach, then at each stage, we make a recursive call. That means leaving the current invocation on the stack and calling a new one. When we are K levels deep, we have got k lots of stack frame, so the space complexity ends up being proportional to the depth you have to search. Space Complexity O(log n).

Sumesh-H:
Good Job, Balaviknesh. Kindly check inline comments.

Sumesh-H:
> I dont have leet code premium account for problem 2. Please let me know how I can get the problem. thanks.

You can implement it in any IDE or GeeksforGeeks.

PR #247, opened 2020-08-19
Completed BinarySearch Problem-1

chinmayeetapaskar:


Gupta-Keshav-001378461:
Good Job!! Submit the second question as well.

PR #261, opened 2020-08-20
Completed Binary Search

manicha321:


khushal1198:
Correct implementation of Rotated array problem. Mention time and space complexity along with comments.

In infinite array problem, you are searching in whole integer range. Kindly try implementing again after you get done with the problem in lecture.

PR #270, opened 2020-08-20
Binary search 1 completed - Please review

dcarao:


Himanshu-Agarwal93:
Great Work, Dcarao

Himanshu-Agarwal93:
Correct implementation for search in a rotated sorted array.
However, you cannot extend the same solution for the second problem. The length of the array could be infinite.

PR #272, opened 2020-08-20
Day 3 prolblems

sidhs5695:


khushal1198:
Good job Siddharth. Correct time and space complexities.

khushal1198:
You have made a mistake for finding the middle element in binary search. It should be l+(h-l)/2 rather than l+(h-1)/2. That should solve your problem

PR #273, opened 2020-08-20
Binary search-1 problem implementation

Jayesh-sathe:
Problem-2 is not present on leetcode.


Gupta-Keshav-001378461:
Nice Work!!

PR #274, opened 2020-08-20
Completed Binary-Search-1, please review.

vatsaashwin:


Himanshu-Agarwal93:
Good job. Second solution will not work for an array of infinite length. How can you improve?

PR #289, opened 2020-08-21
Completed Binary-Search-1(Python) - Please Review

mrithunjaybr:


Himanshu-Agarwal93:
Incorrect approach for serachInArrayOfUnknownSize. What will happen if length of the array is infinite?
Rest is correct.

PR #317, opened 2020-09-17
Problem1 Solution - TLE - Please Review

AkashBudholia:


Problem1.java
```diff
@@ -0,0 +1,85 @@
+// Time Complexity : O(logn)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Time Limit Exceeded
+// Any problem you faced while coding this : How to get the pivot in LogN time
+
+
+// Your code here along with comments explaining your approach
+
```
Lohith-Aswathappa, line 8: You are trying to do too many operations to find the target value.
As taught in class, watch for on which side of the mid you have your target value and process with binary search further.

Lohith-Aswathappa (commented):
Nice try, Akash!
Review the inline comment and try the other problem as well.

Lohith-Aswathappa:
Nice try, Akash!
Review the inline comment and try the other problem as well.

PR #318, opened 2020-09-17

krunalT21:


Solution1.java
```diff
@@ -0,0 +1,34 @@
+class Solution1 {
+    // Time :  O(logn)  Space : O(1)
+    public int search(int[] nums, int target) {
+
```
kannan-mrugapphan, line 4: Space is not constant as you are recursing

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

Solution1.java
krunalT21: Thanks, Kannan I might need to read more about the complexity

PR #326, opened 2020-09-17
init commit for bs1

amitsingh5949:


_33_Search_in_Rotated_Sorted_Array.java
```diff
@@ -0,0 +1,148 @@
+package com.java.binaryserach.leetcode;
+
+import java.util.Scanner;
```
Lohith-Aswathappa, line 3: Try to submit the problem on leetcode and check if the solution is generalized

Lohith-Aswathappa (commented):
Nice try, Amit!
Check the inline comment

Lohith-Aswathappa:
Nice try, Amit!
Check the inline comment

_33_Search_in_Rotated_Sorted_Array.java
amitsingh5949: I did, the package is mentioned for my eclipse workspace i did submit it without this line. Thank you for looking into this :)

PR #338, opened 2020-10-14
Submission

Amrutha2509:
Hi,

For the second problem, I am getting time-limit exceeded. But wanted to know how I can make it efficient or what I might be missing.

Thanks!

### DIFF
--- a/Problem-1.py
+++ b/Problem-1.py
@@ -1,0 +1,23 @@
+class Solution:
+    def search(self, nums: List[int], target: int) -> int:
+
+        #we will use binary search to solve this.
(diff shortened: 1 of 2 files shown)


khushal1198 (commented):
Good work Amrutha. You had made a mistake in the ending range. Make end = (end * 2) and your program will work just fine!

khushal1198:
Good work Amrutha. You had made a mistake in the ending range. Make end = (end * 2) and your program will work just fine!

Amrutha2509:
ok got it thank you!

PR #360, opened 2020-10-16
please review

Mohit0993:


samarthjj:
Good work Mohit, Please also add a brief description of your approach in 3-4 points for every problem.

samarthjj:
Good work Mohit, Please also add a brief description of your approach in 3-4 points for every problem.

PR #379, opened 2020-11-21
Binary search 1 Java implementation

ckallianpur:


Solution.java
```diff
@@ -0,0 +1,51 @@
+/*MyHashSet Implementation
+Runtime is O(logn)
+Space complexity is O(1). No additional space is used
```
samarthjj, line 3: O(log(n)), recursive stack space.

Solution.java
```diff
@@ -0,0 +1,51 @@
...
+The program took a long time to give the answer(7s) 
+Stats from leetcode - Is this increase because of using recursion?
```
samarthjj, line 5: Could be yes, but do not worry about it.
As long as the Asymptotic Time Complexity is log(n).

samarthjj:
Good work Chaitra, Please review the inline comments.
Please also complete the other problem as well.

samarthjj:
Good work Chaitra, Please review the inline comments.
Please also complete the other problem as well.

PR #402, opened 2020-12-24
completed assignment

aviveershetty:
Please review solutions


maithrreye:
Good Job ! To your question, Should we look for a modular code during interview? its upto you how you want to design your code at the end of day your code should be efficient and readable. Also your solution to infinite sorted array looks good to me. It is may be the way leetcode text different from how you test in local as it involves interfaces

maithrreye:
Good Job ! To your question, Should we look for a modular code during interview? its upto you how you want to design your code at the end of day your code should be efficient and readable. Also your solution to infinite sorted array looks good to me. It is may be the way leetcode text different from how you test in local as it involves interfaces

PR #445, opened 2020-12-29
completed BinarySearch-1, Please review

shantanuneema:
Please pay attention to alternative solution for 2D matrix and share your comments

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,103 @@
+# Search in a 2D matrix:
+
+class Solution:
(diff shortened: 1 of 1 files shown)


khushal1198 (commented):
Good work Shantnu. On line 32 make row = mid/n in place of mid/m and your solution will work. For this question, after a couple of weeks, you'll be taught a better solution.

khushal1198:
Good work Shantnu. On line 32 make row = mid/n in place of mid/m and your solution will work. For this question, after a couple of weeks, you'll be taught a better solution.

PR #462, opened 2021-01-20
Implemented BinarySearch-1 problems

itselavia:


Gupta-Keshav-001378461 (commented):
Good Work, Akshay!! But we don't say Space complexity is 0 instead we say it's O(1).

Gupta-Keshav-001378461:
Good Work, Akshay!! But we don't say Space complexity is 0 instead we say it's O(1).

PR #496, opened 2021-02-18
BinarySearch-1 first two problems

dkeerthireddy:


SearchIn2DMatrix.java
```diff
@@ -0,0 +1,102 @@
+// Time Complexity : O(log(n))
```
sibalnirbhay, line 1: It's still O(log mn) (same as your other solution). Notice that your `high` is `m*n`.

sibalnirbhay (commented):
Looks good!
Read inline comments.

sibalnirbhay:
Looks good!
Read inline comments.

PR #538, opened 2021-03-20
Completed Sprint BinarySearch1

yash2396:


khushal1198 (commented):
Great job with the first problem. Please add complexities and your approach in other problems as well.

khushal1198:
Great job with the first problem. Please add complexities and your approach in other problems as well.

PR #594, opened 2021-05-27
Completed Binary-Search-1 v1 - Please Review

isdarwhekar:


kannan-mrugapphan (commented):
Good Job. Try to incorporate a try-catch block to avoid out of bounds exception for the infinite array problem. I would recommend doing binary search iteratively thus saving on call stack

kannan-mrugapphan:
Good Job. Try to incorporate a try-catch block to avoid out of bounds exception for the infinite array problem. I would recommend doing binary search iteratively thus saving on call stack

PR #598, opened 2021-05-27
Completed binary search 1

tanreetD:


maithrreye:
Good Job on Brute Force solution ! you can add TC and SC . Please add the optimal solution

maithrreye:
Good Job on Brute Force solution ! you can add TC and SC . Please add the optimal solution.

PR #609, opened 2021-05-28
Submitting 1st question as of now, please review

SheetalMore9878:
Working on 2nd and 3rd question still, will submit in a while


ramya-sruthi-d (commented):
The approach and code is good, mention the complexities of your approach as well in the comments.

ramya-sruthi-d:
The approach and code is good, mention the complexities of your approach as well in the comments.

PR #660, opened 2021-06-02
Binary Search - 1

hharshitarora:
For the first one I was having a lot of problems in applying the BS on column then row, would really want some feedback on it.
For second I did have to re understand the approach using geeks for geeks
For third I very well understood the idea of expanding search space and applying BS onto it. Though on leetcode speed seemed slow is that a problem ? (Solution is O(logn))

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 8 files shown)


Rooppesh:
For the first problem, the matrix is sorted in the row order, so you don't need to do it along the column. For the third one, the TC would be O(logn). You don't need to worry about the LC speed. Let me know if you have further questions.

Rooppesh:
For the first problem, the matrix is sorted in the row order, so you don't need to do it along the column. For the third one, the TC would be O(logn). You don't need to worry about the LC speed. Let me know if you have further questions.

PR #676, opened 2021-06-24
Completed Binary-Search1(Java) Please Review

sairamya27:


super30admin:
Solution is correct. Add TC and SC

PR #684, opened 2021-06-25
Finished Binary search except 3rd problem

sumeetgupta9696:


Lohith-Aswathappa (commented):
Good work, Sumeet!
Try the optimized binary search approach for both problems

Lohith-Aswathappa:
Good work, Sumeet!
Try the optimized binary search approach for both problems

PR #704, opened 2021-07-22
Binary Search 1

harshasatya:
Done all 3 questions


Rooppesh:
Good Job! Please add the TC and SC as well!

Rooppesh:
Good Job! Please add the TC and SC as well!

PR #705, opened 2021-07-22
completed binary search 1,

yashnaredi:
solutions avaialble both in python and java, I am still getting used to java syntax , so i rewrote everything in both languages.


rimzimt (commented):
Great effort Yash with all the solutions!

rimzimt:
Great effort Yash with all the solutions!

PR #752, opened 2021-08-19
Binary Search-1 completed(Ex-1&2), Please review

shilpagowda812:


sridhar-anish:
Good Job with everything, make sure to write TC and SC also

PR #803, opened 2021-09-16
Complete BinarySearch-1, Please review

anandreddysripathi:


maithrreye:
I guess your first approach will work if you simply start low=1 and high=low*2 may be.But good job with the other version !

maithrreye:
I guess your first approach will work if you simply start low=1 and high=low*2 may be.But good job with the other version !

PR #809, opened 2021-09-20
Completed sorted rotated array & 2D matrix - Please review

vishwaparikh:
Will completed 3rd one later didnt undertand the size of the end how to determine . Could not do it.

### DIFF
--- a/SearchIn2DMatrix
+++ b/SearchIn2DMatrix
@@ -1,0 +1,34 @@
+class Solution {
+    public boolean searchMatrix(int[][] matrix, int target) {
+        int m = matrix.length; //row length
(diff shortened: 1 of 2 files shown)


Pinak08 (commented):
Good Job!
One more way to not make it more complex in problem 1 is to traverse row separately using binary search and column separately using another binary search so it can be done in O(Logn)*O(logm).

Pinak08:
Good Job!
One more way to not make it more complex in problem 1 is to traverse row separately using binary search and column separately using another binary search so it can be done in O((Logn)(logm)).

vishwaparikh:
Yes @Pinak08  Thank you for suggesting me I will that as well.

PR #820, opened 2021-09-23
completed Binary Search-1 - Please Review

niharikagupta2021:


manojvarmarudraraju (commented):
Good job, please include the structure as well!!

manojvarmarudraraju:
Good job, please include the structure as well!!

unknown_array_size.py
```diff
@@ -1,3 +1,10 @@
+"""
```
manojvarmarudraraju, line 1: As we are iterating through unlimited size array our approach is similar O(infinity) in linear search but in our approach it will be O(log(infinity))

manojvarmarudraraju (commented):
Good job!!!

PR #847, opened 2021-10-14
Completed Sprint BS1 - Please review

joshyoon92:


Search2D.java
```diff
@@ -0,0 +1,25 @@
+
```
sanjayvarma199, line 1: class Solution {
    //Time O(m+n)
    //Space O(1)
        if(matrix == null || matrix.length == 0)
        {
            return false;
        }
        int row = matrix.length-1 , col = 0;
        while(row >=0 && col < matrix[0].length)
        {
            if(matrix[row][col] == target)
            {
                return true;
            }
            else if(matrix[row][col] > target)
            {
                row--; // All the numbers to its clearly greater than the current element so move upwards
            }
            else
            {
                col++;
            }
        }
        return false;
    }
}

PR #907, opened 2021-12-23
Completed Binary-Search-1 : Please review

ravirammamidi:


sanjayvarma199 (commented):
Good Job!! For search 2D matrix You can perform the binary search in single go rather than performing on row and then on column.

sanjayvarma199:
Good Job!! For search 2D matrix You can perform the binary search in single go rather than performing on row and then on column.

PR #914, opened 2021-12-24
Completed BinarySearch-1, please review. 

sd7297:


sonalgupta276 (commented):
TC and SC looks wrong. Also complete the other problem. Kindly update after you attend the class for these problems

sonalgupta276:
TC and SC looks wrong. Also complete the other problem. Kindly update after you attend the class for these problems

PR #923, opened 2021-12-24
Completed Binary Search 1 problems

swaroopguptaba:


sonalgupta276 (commented):
Good job! Mention TC and SC as well

sonalgupta276:
Good job! Mention TC and SC as well

PR #926, opened 2021-12-24
Binary Search- Please Review

manchilis:


Gupta-Keshav-001378461:
Codes are working but this is not an ideal submission. Ideal submission consists of all these:
 // Time Complexity : you mentioned this good.
// Space Complexity : you mentioned this good.
// Did this code successfully run on Leetcode : Mention this as well
// Any problem you faced while coding this : Mention this as well it's a simple yes and no if yes then right the reason to keep track of the problem.

// Your code here along with comments explaining your approach
The code with comments is the most important for example if you come back after 4 or 5 months to review the solution you don't have to go through the code line by line to remember what you have done if you right meaningful comments. 

And one more thing to make the code for unknown size a little faster you can use right shift operation instead of multiplication.

Gupta-Keshav-001378461:
Codes are working but this is not an ideal submission. Ideal submission consists of all these:
 // Time Complexity : you mentioned this good.
// Space Complexity : you mentioned this good.
// Did this code successfully run on Leetcode : Mention this as well
// Any problem you faced while coding this : Mention this as well it's a simple yes and no if yes then right the reason to keep track of the problem.

// Your code here along with comments explaining your approach
The code with comments is the most important for example if you come back after 4 or 5 months to review the solution you don't have to go through the code line by line to remember what you have done if you right meaningful comments. 

And one more thing to make the code for unknown size a little faster you can use right shift operation instead of multiplication.

PR #930, opened 2021-12-25
Completed Problem 2 and 3 from BinarySearch-1

anmols97:


sanjayvarma199 (commented):
Good Job!! for Problem1 TC is log(m*n) not log(m+n)

sanjayvarma199:
Good Job!! for Problem1 TC is log(m*n) not log(m+n)

PR #951, opened 2022-01-01
python solution

gdg4dev:


sonalgupta276 (commented):
Good job! Mention TC and SC for last problem

sonalgupta276:
Good job! Mention TC and SC for last problem

PR #961, opened 2022-01-10
Completed Binary Search-1

ashwinsaxena47:
Completed Binary Search-1
1. Search 2d matrix
2. Search inside rotated sorted array
3. Search inside a sorted array of unknown length

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,53 @@
+/* Problem 2: Search inside rotated sorted array
+*/
+// Time Complexity : O(log n)
+// Space Complexity : O(1) since we are not using any auxiliary space
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+
+// Your code here along with comments explaining your approach
+//Follow up question(interviewer): Can we do this in single pass?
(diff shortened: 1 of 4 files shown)


Problem2.java
```diff
@@ -0,0 +1,53 @@
+/* Problem 2: Search inside rotated sorted array
+*/
+// Time Complexity : O(log n)
+// Space Complexity : O(1) since we are not using any auxiliary space
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+
+// Your code here along with comments explaining your approach
+//Follow up question(interviewer): Can we do this in single pass?
```
manojvarmarudraraju, line 10: We are doing in single pass only, We will not touch every element but atmost we will touch only once. So it is single pass algorithm only

manojvarmarudraraju (commented):
Good job, Nice explanations keep it going!!!!

manojvarmarudraraju:
Good job, Nice explanations keep it going!!!!

Problem2.java
ashwinsaxena47: Okay thanks for your feedback.

PR #1041, opened 2022-04-01
Exercise1

rujal8452:


sanjayvarma199 (commented):
Good Job!! Please specify the TC and SC

sanjayvarma199:
Good Job!! Please specify the TC and SC

PR #1049, opened 2022-04-01
Binary Search Problem 1 and 2: Please Review had some issues in Problem2

shrey24:
Problem2 see `Any problem you faced while coding this ` comment

### DIFF
--- a/problem_1.py
+++ b/problem_1.py
@@ -1,0 +1,29 @@
+""" Search in 2D Matrix """
+# Time Complexity : O(Log(m*n)) where m: #rows and n: #cols
+# Space Complexity : O(1)
+# Did this code successfully run on Leetcode : Yes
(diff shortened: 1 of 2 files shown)


keshav-Gupta-IS (commented):
Good Job, Shrey!! Your thought process for problem 2 will fail for cases like 
[1,10,11]
[2,  3,  5]
[4, 12, 13]
target = 3 Here how would you decide which row to select? You can elaborate if you think something is missing.

keshav-Gupta-IS:
Good Job, Shrey!! Your thought process for problem 2 will fail for cases like 
[1,10,11]
[2,  3,  5]
[4, 12, 13]
target = 3 Here how would you decide which row to select? You can elaborate if you think something is missing.

PR #1094, opened 2022-04-28
Completed Sprint 2D Array. Please review. Thank you

divya-kulkarni99:
I tried to solve the other two problem but did not get the solution.

### DIFF
--- a/Sample
+++ b/Sample
@@ -1,7 +1,59 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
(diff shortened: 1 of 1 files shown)


manojvarmarudraraju (commented):
Good job, try the brute-force solution. The optimal solution will be taught in the class!!

manojvarmarudraraju:
Good job, try the brute-force solution. The optimal solution will be taught in the class!!

divya-kulkarni99:
Thank you for the review. I will try doing brute force if I did do the optimal solution from now on.

PR #1145, opened 2022-05-26
2dmatrix- please review

HanumanReddy1212:


Sakshi-2103 (commented):
Good job
Try other problems as well
Add TC and SC to all the problems

Sakshi-2103:
Good job
Try other problems as well
Add TC and SC to all the problems

PR #1150, opened 2022-05-26
both 5 and 6 done please review

sai-pavan-kalyan:


keshav-Gupta-IS (commented):
Difficulty in determining the if condition reason might be you are not dry running these problems on whiteboard or paper with different examples that's the way to come up to the solution. Also it is the beginning of cohort you will solve such problems in no time moving forward

keshav-Gupta-IS:
Difficulty in determining the if condition reason might be you are not dry running these problems on whiteboard or paper with different examples that's the way to come up to the solution. Also it is the beginning of cohort you will solve such problems in no time moving forward

PR #1230, opened 2022-06-24
Exercise 3 remaining as not taught in class. Will complete it after t…

viveksharma180:
Exercise 3 remains as not taught in class. Will complete it after tomorrow's class.

Signed-off-by: viveksharma180 <sharma.viv@northeastern.edu>

### DIFF
--- a/rotatedSortedArray.java
+++ b/rotatedSortedArray.java
@@ -1,0 +1,40 @@
+// Time Complexity : O(logN)
+// Space Complexity : O(1) constant
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : understanding logic inside first else condition. It could have been
+//   if(target > nums[low] && target <= nums[mid]){
+//                    high = mid-1;
+//                }else{
+//                    low=mid+1;
+//                }
+// But it is not working properly
+
+// Your code here along with comments explaining your approach
+
+class Solution {
+    public int search(int[] nums, int target) {
+        int low = 0;
+        int high = nums.length -1;
+
+        while(low <= high){
+            int mid = low + (high-low)/2;
+            if(target == nums[mid]){
+                return mid;
+            }
+            if(nums[low] <= nums[mid]){ // going in sorted array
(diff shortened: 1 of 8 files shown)


rotatedSortedArray.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(logN)
+// Space Complexity : O(1) constant
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : understanding logic inside first else condition. It could have been
+//   if(target > nums[low] && target <= nums[mid]){
```
jeelpatel18, line 5: See, if target is at mid we are returning mid, but what if target > nums[low] || target is at nums[low] itself, in that case we are searching towards left right. or target is lesser than element at mid

rotatedSortedArray.java
```diff
@@ -0,0 +1,40 @@
...
+// Your code here along with comments explaining your approach
+
+class Solution {
+    public int search(int[] nums, int target) {
+        int low = 0;
+        int high = nums.length -1;
+
+        while(low <= high){
+            int mid = low + (high-low)/2;
+            if(target == nums[mid]){
```
jeelpatel18, line 21: If we get target at mid, we don't need to search anywhere else then

rotatedSortedArray.java
```diff
@@ -0,0 +1,40 @@
...
+                return mid;
+            }
+            if(nums[low] <= nums[mid]){ // going in sorted array
+                if(target >= nums[low] && target < nums[mid]){ // if true then it means target lies in left side
+                    high = mid-1;
+                }else{
+                    low = mid +1; // If not, then in right side
+                }
+            }else{
+                if(target > nums[mid] && target <= nums[high]){ // searching here and also because at last now it is sorted part
```
jeelpatel18, line 31: Same for right side also

jeelpatel18 (commented):
Good job Vivek. Do a PR when you submit the 3rd problem.

jeelpatel18:
Good job Vivek. Do a PR when you submit the 3rd problem.

PR #1283, opened 2022-07-21
completed_sprint...please review

Sreelekhasamanu:


daredevilteja (changes requested):
For 2D matrix this is the solution by modifying your code
```

class Solution {
        int m = matrix.length;
        System.out.println(m);
        int n = matrix[0].length;
       
        int i=0;
        int low = 0;
        int high = n;
        int value=-1;
        while(i<m){
            if(target <= matrix[i][n-1] && matrix[i][0] <= target)
            {value =  BS(matrix,0,high,target,i);
            break;
            }
            i++;
        }
        if(value==-1)
            return false;
        return true;

    }

    public int BS(int[][] matrix, int low, int high, int target, int row){ 
        while(low <= high) {
            int mid = (low + high) / 2;
            if (target == matrix[row][mid])
                return mid;
            else if (target > matrix[row][mid])
                low = mid + 1;
            else
                high = mid - 1;
        }
        
        return -1;
        
    }
}


```but the time complexity will be O(rows * log columns) . you have to still modify the solution to get log(rows*columns)

daredevilteja:
Please check the comments. Good Work! Keep it up

PR #1324, opened 2022-08-21
Completed Binary-Search-2: Please Review

jpspecter:


daredevilteja:
Please mention TC, SC and whether it executed on leetcode for every problem

PR #1347, opened 2022-08-29
Completed - Please Review

sid-pandya:


daredevilteja (changes requested):
Search in infinite sorted array is wrong.
This is the solution 
```
class Solution:
        low = 0
        high = 1
        while(reader.get(high) < target):
            low = high
            high = 2 ** high
        while(low <= high):
            mid = low + (high - low) // 2
            if (reader.get(mid) == target):
                return mid
            elif reader.get(mid) > target:
                high = mid - 1
            else:
                low = mid + 1
        return -1 
```

daredevilteja (changes requested):
Search in infinite sorted array is wrong.
This is the solution 
```
class Solution:
        low = 0
        high = 1
        while(reader.get(high) < target):
            low = high
            high = 2 ** high
        while(low <= high):
            mid = low + (high - low) // 2
            if (reader.get(mid) == target):
                return mid
            elif reader.get(mid) > target:
                high = mid - 1
            else:
                low = mid + 1
        return -1 
```

daredevilteja:
Please do the requested changes

sid-pandya:
Please Review these points for further clarification.
- This is LC Premium problem so haven't checked the test cases in LC 
- I can not implement 'ArrayReader' in local environment.
-  "high = 2 ** high " in your code implements 2 to the power of high which is not correct in my opinion.
Please let me know the next steps how to correct my mistakes if there's any.
Thanks.

daredevilteja:
> Please Review these points for further clarification.
> 
> * This is LC Premium problem so haven't checked the test cases in LC
> * I can not implement 'ArrayReader' in local environment.
> * "high = 2 ** high " in your code implements 2 to the power of high which is not correct in my opinion.
>   Please let me know the next steps how to correct my mistakes if there's any.
>   Thanks.

1. What is wrong with high = 2 ** high in your opinion? 
2. You don't need to implement ArrayReader, it is already implemented. You have to use the method given to you.

PR #1349, opened 2022-08-30
Completed Binary-Search-1 Please review

SnehaNellutla:


daredevilteja (commented):
SerachInfinitelength solution is wrong. This is the solution
```
class Solution {
    public int search(ArrayReader reader, int target) {
        int i = 0, low = 0, high = i, mid;
        while(target >= reader.get(high)) {
            low = high;
            high = (int)Math.pow(2,high);
        }
      
        while(high >= low) {
            mid = low + (high - low)/2;
            
            if(reader.get(mid) < target) {
                low = mid + 1;
            } else if(reader.get(mid) > target) {
                high = mid - 1;
            } else {
                return mid;
            }
        }
        return -1;
    }
}
```

daredevilteja:
Good Work! Keep it up

PR #1360, opened 2022-09-09
Binary search 1 solutions

kundankarthik77:


jeelpatel18 (commented):
Good job. Please add TC and SC and 3rd problem too.

jeelpatel18:
Good job. Please add TC and SC and 3rd problem too.

PR #1367, opened 2022-09-09

sh-ruchi:


daredevilteja:
Please complete all the problems and then raise the PR

sh-ruchi:
@daredevilteja  Added the solution . wrongly  Added empty file

daredevilteja:
> @daredevilteja Added the solution . wrongly Added empty file

There is another problem searchinfinite array

sh-ruchi:
@daredevilteja  done

SearchInSorterArrUnknownSize.java
```diff
@@ -0,0 +1,32 @@
+// Time Compexity : O(log n)
```
daredevilteja, line 1: What is n?

daredevilteja:
Good Work! Keep it up

Search2dMatrix.java
```diff
@@ -0,0 +1,38 @@
+// Time complexity is: O(Log n)
```
daredevilteja, line 1: TC is wrong

SearchRotatedArray.java
```diff
@@ -0,0 +1,43 @@
+// Time complexity : O(log n)
```
daredevilteja, line 1: what is n?

daredevilteja (changes requested):

daredevilteja:
Please do the requested changes

SearchInSorterArrUnknownSize.java
sh-ruchi: it should actually be O (Log mn)

sh-ruchi: O(log n) in range high- low

Search2dMatrix.java
sh-ruchi: it should be O (Log mn) because its a 2D matrix

SearchRotatedArray.java
sh-ruchi: n is size of the array and we are diving the array to search and element in one half of the array .
Is my understanding correct ?

sh-ruchi:
done

SearchInSorterArrUnknownSize.java
daredevilteja: every time you have to mention the variables you are using for your complexities, here n is not present in code and you have mentioned it in TC so you have to explicitly tell what n is. 
For example here 
TC is O(log n) where n is value of the high variable

SearchRotatedArray.java
daredevilteja: you have to mention like 
TC is O(log n) where n is the length of the nums array

PR #1392, opened 2022-09-11
Please Review

rashidhir6:


Problem 3
```diff
@@ -0,0 +1,39 @@
+// Time Complexity : O(log T )
```
daredevilteja, line 1: what is T

Problem 1
```diff
@@ -1,4 +1,4 @@
-// Time Complexity : O(n)
+// Time Complexity : O(log n)
```
daredevilteja, line 1: TC is O(log mn)

daredevilteja:
Good Work! Keep it up

PR #1426, opened 2022-10-10
Binary search1 done

tanaytadas16:


2D_Matrix.java
```diff
@@ -0,0 +1,27 @@
+//time complexity is : 0(logn) and space complexity is 0(1)
```
daredevilteja, line 1: TC is O(log(totalRow*totalCol))

daredevilteja (changes requested):
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

daredevilteja:
Please do requested changes

PR #1429, opened 2022-10-11
Binary-search:1 Please Review

pavansai1:


BS2d.java
```diff
@@ -0,0 +1,33 @@
+// Time Complexity : O(logn)
```
daredevilteja, line 1: TC is O(log(m*n))

BSInfiniteSorted.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(log n)
```
daredevilteja, line 1: TC is O(m/2 + logn) according to your implementation

daredevilteja:
Good Work! Keep it up

PR #1496, opened 2022-12-25
BS-1 Please Review

Rohan-Balakrishna:


daredevilteja (changes requested):
Resubmit Search_Unknown_length problem. There is no relation between both functions in the problem you submitted.

```
# """
# This is ArrayReader's API interface.
# You should not implement it, or speculate about its implementation
# """
#class ArrayReader:
#    def get(self, index: int) -> int:

class Solution:
        ```

daredevilteja:
Good Work! Keep it up. Please make the requested changes.

PR #1538, opened 2023-03-01
added binary search-1 solution here

akhilgour:


akhilgour:
added solution-searching in infinite sorted array

niharikagupta2021 (commented):
Good Job!

PR #1557, opened 2023-03-17
Completed Binary Search-1.

AnirudhS30:


Rohan-Balakrishna:
Good Job Pls mention TC and SC as well.

Rohan-Balakrishna:
Good Job Pls mention TC and SC as well.

PR #1633, opened 2023-06-30
Completed Brute Force approach

SudhaRamachandran:


searchMatrix.py
```diff
@@ -0,0 +1,29 @@
+# Brute Force Solution
```
ishika2741, line 1: Submit the optimised solution for this problem where you do not need to copy the matrix elements in new list

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1696, opened 2023-08-21
Binary-seach-1 all

ghost:
please review


Nandha2Tech (commented):
Great Work!
For Array Reader TC would be  O(log n + log m) setting the search range takes O(log n) & searching in the search range takes O(log m). SC would be O(1).

Nandha2Tech:
Great Work!
For Array Reader TC would be  O(log n + log m) setting the search range takes O(log n) & searching in the search range takes O(log m). SC would be O(1).

PR #1708, opened 2023-09-01
Binary Search 1

UmangThakur15:
Please Review


Nandha2Tech (commented):
Great Work!
SC for Rotated_Sorted_Array_Search would be O(1).
Kindly complete search in a sorted array of unknown size problem also.

Nandha2Tech:
Great Work!
SC for Rotated_Sorted_Array_Search would be O(1).
Kindly complete search in a sorted array of unknown size problem also.

UmangThakur15:
That question is not visible as it is only for premium members and I don't have premium. Can you please share the complete problem with me?

Nandha2Tech:
Hi, I think that problem is not completed yet. It will be completed in your next session. So no problem. You can submit it once it is completed.

UmangThakur15:
I have implemented the Search in Infinite Array as well. Please review.

Nandha2Tech (commented):
Great Work!

PR #1719, opened 2023-09-02
Completed Binary-Search-1 - Kindly Review.

apurvashinde357:


Nandha2Tech (commented):
Great Work!
Since we are doing binary search TC would be O(log n) 
Please submit another problem also once completed.

Nandha2Tech:
Great Work!
Since we are doing binary search TC would be O(log n)
Please submit another problem also once completed.

PR #1740, opened 2023-09-29
Completed Binary-Search-1

Mohini-Mayekar:
Problem #6 - search in rotated sorted array
Problem #7 - search in a sorted array of unknown size
Problem #8 - 2D matrix

### DIFF
--- a/Search_SortedArr_UnknownSize.java
+++ b/Search_SortedArr_UnknownSize.java
@@ -1,0 +1,29 @@
+// Time Complexity : O(log (m*n))
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : -
+// Any problem you faced while coding this : -
+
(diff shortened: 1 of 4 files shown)


Search_SortedArr_UnknownSize.java
```diff
@@ -0,0 +1,29 @@
+// Time Complexity : O(log (m*n))
```
ishika2741, line 1: Kindly explain what m and here is

ishika2741 (commented):
Good work!

ishika2741:
Good work!

Search_SortedArr_UnknownSize.java
Mohini-Mayekar: m - to find the range (low and high) - first while loop
n - the binary search - second while loop

PR #1742, opened 2023-09-29
Create Binary_Search_Rotated_Array.py

KrishanuSinha:
There is an integer array nums sorted in ascending order (with distinct values).

Prior to being passed to your function, nums is possibly rotated at an unknown pivot index k (1 <= k < nums.length) such that the resulting array is [nums[k], nums[k+1], ..., nums[n-1], nums[0], nums[1], ..., nums[k-1]] (0-indexed). For example, [0,1,2,4,5,6,7] might be rotated at pivot index 3 and become [4,5,6,7,0,1,2].

Given the array nums after the possible rotation and an integer target, return the index of target if it is in nums, or -1 if it is not in nums.

You must write an algorithm with O(log n) runtime complexity.

### DIFF
--- a/Binary_Search_Rotated_Array.py
+++ b/Binary_Search_Rotated_Array.py
@@ -1,0 +1,24 @@
+class Binary_Search_Rotated_Array:
+    def search(num:list[int],target:int) -> int:
+        l,r =0, len(num) -1
+        while l <=r:
+            mid = (l+r)//2
+            
+            if target == num[mid]:
+                return mid
+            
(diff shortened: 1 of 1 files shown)


keshav-Gupta-IS (commented):
Good Job, Krishanu!! Keep it up. Mention time and space complexities.

keshav-Gupta-IS:
Good Job, Krishanu!! Keep it up. Mention time and space complexities.

PR #1744, opened 2023-09-29
Create Binary_Search.py

KrishanuSinha:
This is an interactive problem.

You have a sorted array of unique elements and an unknown size. You do not have an access to the array but you can use the ArrayReader interface to access it. You can call ArrayReader.get(i) that:

returns the value at the ith index (0-indexed) of the secret array (i.e., secret[i]), or returns 231 - 1 if the i is out of the boundary of the array. You are also given an integer target.

Return the index k of the hidden array where secret[k] == target or return -1 otherwise.

You must write an algorithm with O(log n) runtime complexity.

### DIFF
--- a/Binary_Search.py
+++ b/Binary_Search.py
@@ -1,0 +1,18 @@
+class Binary_Search:
+    def Solution(num:list[int],target:int)->int:
+        l,r = 0,len(num)-1
+        
+        while (l<=r):
+            mid = (l+r)//2
+            
+            if target == num[mid]:
+                return mid
+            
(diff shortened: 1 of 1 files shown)


KrishanuSinha:
Merge

keshav-Gupta-IS (commented):
Good Job, Krishanu!! Keep it up. Mention time and space complexities.

keshav-Gupta-IS:
Good Job, Krishanu!! Keep it up. Mention time and space complexities.

PR #1757, opened 2023-09-30
Binary Search -1

ArchitShahMartian:


Nandha2Tech (commented):
Great Work!
For problem 1, search in a 2d matrix, TC would be log m + log n. Since to identify the row binary search is done which is log m and to identify the target in that row binary search takes log n. So total TC would be log m + log n.

Nandha2Tech (commented):
Great Work!
For problem 1, search in a 2d matrix, TC would be log m + log n. Since to identify the row binary search is done which is log m and to identify the target in that row binary search takes log n. So total TC would be log m + log n.

Nandha2Tech:
Great Work!
For problem 1, search in a 2d matrix, TC would be log m + log n. Since to identify the row binary search is done which is log m and to identify the target in that row binary search takes log n. So total TC would be log m + log n.

PR #1780, opened 2023-11-11
Completed Binary Search 1

rajathbrgowda:


Nandha2Tech (commented):
Great Work!
For search in a 2D matrix TC would O(log m*n)
For search in an infinite array low pointer can also be updated to reduce the search range.

Nandha2Tech:
Great Work!
For search in a 2D matrix TC would O(log m*n)
For search in an infinite array low pointer can also be updated to reduce the search range.

PR #1820, opened 2023-12-26
added two files

Sreedharreddy20:


vyankateshmisar:
Good Job Sreedhar. Please include TC and SC in every problem.

vyankateshmisar:
Good Job Sreedhar. Please include TC and SC in every problem.

PR #1935, opened 2024-06-20
completed binary search 1 homework

dhikki-study:
##############rotated sorted array ####################
// Time Complexity : should be O(n) as I used loop to identify place of pivot
// Space Complexity : not sure
// Did this code successfully run on Leetcode : following case failed and other may fail
[1,3]
// Any problem you faced while coding this : made a mistake to take mid for assigning to left and right pointer but it should be mid+1 and mid-1. Also, for running while loop for BS I took < instead of <=

##############702. Search in a Sorted Array of Unknown Size####################
// Time Complexity : logn
// Space Complexity : not sure
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : made a mistake to take mid for assigning to left and right pointer but it should be mid+1 and mid-1. Also, for running while loop for BS I took < instead of <=

// DOUBT - why is r1=10000 working and r1=10**4 not working

##############74. Search a 2D Matrix##########################
// Time Complexity : logm*n
// Space Complexity : not sure
// Did this code successfully run on Leetcode : initial classes passed, time exceeded on following case
[[1,3,5]]
// Any problem you faced while coding this : no


Nandha2Tech (commented):
For SearchInRotatedSortedArray identifying the pivot logic will not work. You are missing the case when the pivot is the last element and also kindly recheck your code. In line no 25 comparision is not proper. Kindly try to verify the entire code and also implement in an optimal manner.
For Search in a sorted array of unknown size r1=10**4 will also work.
For Search in a 2d matrix you did as  coll=mid+1 in line no 119 which should be col1 = mid1+1 and it will work with this change.

Nandha2Tech:
For SearchInRotatedSortedArray identifying the pivot logic will not work. You are missing the case when the pivot is the last element and also kindly recheck your code. In line no 25 comparision is not proper. Kindly try to verify the entire code and also implement in an optimal manner.
For Search in a sorted array of unknown size r1=10**4 will also work.
For Search in a 2d matrix you did as  coll=mid+1 in line no 119 which should be col1 = mid1+1 and it will work with this change.

PR #2014, opened 2024-09-23
implemented questions of binary-1

keyur-patel-21:


Erlybird:
Please add the TC, SC and approach at the heading of the code.

Erlybird:
Please add the TC, SC and approach at the heading of the code.

PR #2041, opened 2024-10-25
Done Binary Search 1

Abhinav-Sriharsha:


Erlybird:
In the third problem there is syntax error for the Array Reader but the logic is correct.

Erlybird:
In the third problem there is syntax error for the Array Reader but the logic is correct.

PR #2107, opened 2025-01-01
Added new file

padmaanaokar:


searchInUnknownSizeRSA.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity :O(log n)
```
ishika2741, line 1: Can you explain your complexity more elaborately. Since there is one while loop for searching the range where target can lie , so even that would contribute to complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2108, opened 2025-01-01
Completed Binary-Search-1

geetanjali-jha:


Search2DMatrix.java
```diff
@@ -0,0 +1,52 @@
+// Time Complexity: O(m log n), where m is the number of rows and n is the size of the column
```
ishika2741, line 1: Can you also submit it with the approach where we assume that we flatten this arr to 1-d arr?

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2165, opened 2025-01-27
Binary-Search-1 completed

amaanJvd:


2DMatrixSearch.java
```diff
@@ -0,0 +1,30 @@
+/* Solved this using imaginary flattened 1D array
+* Time complexity is O(LogN)
```
ishika2741, line 2: Complexity should be in terms of m and n since it as a matrix

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2322, opened 2025-06-22
completed binary-search1

JoshnaRavi04:


super30admin:
- For the rotated sorted array problem, the logic for adjusting `low` and `high` is incorrect. The conditions should check if the left or right half is sorted and then determine if the target lies within that sorted half. The current implementation does not correctly handle these cases, leading to incorrect search results.
- The 2D matrix solution is well-implemented and correctly solves the problem. The use of integer division and modulus to map the 1D index to 2D coordinates is a good approach.
- Consider adding more comments to explain the logic, especially for the rotated sorted array problem, to make it clearer why certain conditions are used.

PR #2360, opened 2025-07-29
Janhvi Guha Binary search 1

Janhvi2000:


super30admin:
- Strengths:
  - The solutions handle edge cases and constraints effectively.

- Areas for Improvement:
  - In Problem3.py, the initial approach to find the high bound could be more efficient. Instead of starting with a fixed high value (10000), a better approach would be to exponentially increase the high bound until the target or the array bounds are found.
  - The comments in Problem1.py mention "Time complexity - O(log (m * n)) where m is number of rows and n is number of columns," but the problem description is not provided. It would be helpful to clarify the problem context in the comments.

PR #2371, opened 2025-08-21
RoatedSortedArray & UnknownSizeSorted array Search: Done

hp6318:


super30admin:
Strengths:
- The comments and explanations provided are clear and helpful for understanding the logic.
- The code is well-organized and follows best practices.

Areas for Improvement:
- In the first problem (search in rotated sorted array), the condition `if target<nums[mid] and target>=nums[left]` could be written more clearly as `if nums[left] <= target < nums[mid]` to match the reference solution's style.
- In the second problem (search in unknown size sorted array), the student could add a check to handle the case where the target is smaller than the first element of the array, which would make the solution more robust.
- The student could also consider adding more edge case checks, such as handling empty arrays or arrays with a single element, even though the current implementation already handles these cases.

PR #2417, opened 2025-11-14
partially completed bs-1 (unable to find the premium question and 2d wasnt completed in the class, will redo it once done.

PurvaT-11:


super30admin:
- Strengths: The student has correctly implemented the binary search algorithm for a rotated sorted array. The logic is sound, and the code is efficient. The use of comments to explain the approach is good practice.
- Areas for Improvement: The docstring at the top could be expanded to provide a more detailed explanation of the problem and the approach taken. Additionally, the condition `nums[high] >= target` could be written as `nums[high] > target` to be more consistent with the left half check (`nums[mid] > target`), though both are correct.

PR #2428, opened 2025-11-17
Binary Search - 1 solutions

shreyargh:
ℹ️  Does not include `Search a 2D Matrix`

### DIFF
--- a/SearchInRotatedArray.java
+++ b/SearchInRotatedArray.java
@@ -1,0 +1,41 @@
+// Time Complexity : O(log n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The student has a good understanding of binary search and its application in rotated sorted arrays and unknown-sized arrays.
- The time and space complexity are correctly identified and achieved.

Areas for Improvement:
- For "SearchInRotatedArray.java", the condition `nums[mid] > target` could be written as `target < nums[mid]` for better readability and consistency with the reference solution.
- In "SearchInUnknownArray.java", while the solution is correct, it would be beneficial to handle edge cases explicitly, such as when the target is smaller than the first element or larger than the last element in the expanded range.

PR #2437, opened 2025-12-26
Completed BinarySearch1

JainShubham23:


super30admin:
- **Strengths**:
  - The student has a good understanding of binary search and its variations.

- **Areas for Improvement**:
  - For Problem 1, the comment `## important concept here is how we get the row and col for value retrival from matrix` could be more detailed to explain why dividing by `cols` gives the row and modulo by `cols` gives the column.
  - For Problem 2, the comment `## because the array is not sorted, so we first need to find out sorted array` could be clarified to emphasize that the array is rotated, not unsorted, and that one half is always sorted.
  - For Problem 3, the comment `## here we need to do binary search twice` could be more precise by explaining that the first part is to find the bounds and the second part is the actual binary search.

PR #2440, opened 2025-12-26
Completed BinarySearch 1

thondaha:
1. Search an element in rotated sorted array
2. Search an element in an array of unknown size
3. Search an element in a 2D matrix

### DIFF
--- a/src/binarysearch/ArrayReader.java
+++ b/src/binarysearch/ArrayReader.java
@@ -1,0 +1,5 @@
+package binarysearch;
+
+public interface ArrayReader {
+    int get(int index);
+}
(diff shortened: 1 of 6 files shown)


super30admin:
Strengths:
- The student has implemented multiple binary search variations correctly (2D matrix and unknown size array).
- The code is well-commented and follows good practices like using `low + (high - low)/2` to avoid overflow.
- The test cases cover various scenarios, demonstrating good testing practices.

Areas for Improvement:
- The rotated sorted array solution has a logical error in the condition for the right sorted side. The condition `else if(nums[mid] < target)` should be `else` to handle all cases where the right side is sorted.
- There are minor inconsistencies in spacing and formatting that could be improved for better readability.
- The condition `if (reader.get(mid) >= target)` in the unknown size array solution could be simplified to `if (val > target)` since `val == target` is already checked.

PR #2445, opened 2025-12-28
Search a 2D Matrix

Priygupt13:


super30admin:
- The student submitted a solution for the wrong problem ("Search a 2D Matrix") instead of the intended problem ("Search in Rotated Sorted Array").
- The approach and logic are correct for the 2D matrix problem, but they do not apply to the rotated sorted array problem.
- The student should ensure they are solving the correct problem before submitting their solution.
- For the rotated sorted array problem, the student should consider the properties of a rotated sorted array and how binary search can be adapted to handle the rotation.
- The student should also ensure that the method names and variables are correctly spelled (e.g., "binarySerach" should be "binarySearch").

PR #2467, opened 2026-02-07
Add search algorithms for rotated array, 2D matrix, and unknown size …

Balaviknesh:
…array

### DIFF
--- a/RotatedSearchArray.kt
+++ b/RotatedSearchArray.kt
@@ -1,0 +1,33 @@
+//Time complexity: O(log n) where n is the number of elements in the array
+//Space complexity: O(1) as we are using constant space to store the indices
+class RotatedSearchArray {
+    fun search(nums: IntArray, target: Int): Int {
+        var left = 0
+        var right = nums.size - 1
+
+        while (left <= right) {
+            val mid = left + (right - left) / 2
+
+            if (nums[mid] == target) {
+                return mid
+            }
+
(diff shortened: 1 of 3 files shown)


super30admin:
Strengths:
- The solution correctly implements the binary search algorithm for a rotated sorted array.
- The code is clean and well-commented.

Areas for Improvement:
- While the solution is correct, it's important to note that the problem requires a solution for a specific problem. The submission includes multiple classes for different problems. In a real scenario, you should submit only the solution for the problem at hand. However, since this is an evaluation, we are only considering the relevant class.
- The condition in the right sorted part: `if (target > nums[mid] && target <= nums[right])` is correct, but note that the reference solution uses the same logic. However, in Kotlin, we can write it exactly as in the reference.

One minor point: the problem states that the array has distinct values, so we don't have to worry about duplicates. The solution leverages this.

Overall, the solution is excellent.

super30admin:
Your solution for the rotated sorted array search is excellent. It correctly identifies which half of the array is sorted and adjusts the search range accordingly. The time and space complexity are optimal. The code is clean and well-commented. However, please note that when submitting solutions, you should only include the relevant code for the problem at hand. Including unrelated problems (like Search2DMatrix and SearchArrayUnknownSize) in the same file or submission might cause confusion. For future submissions, focus on the specific problem being evaluated.

PR #2469, opened 2026-02-08
S30 FAANMG Problem super30admin#8 | Search inside a 2D Matrix

sandy7907:


super30admin:
Your solution is well-written for the problem of searching in a 2D matrix, but it does not solve the given problem. The problem requires searching in a rotated sorted one-dimensional array. You need to implement a binary search that accounts for the rotation. 

Here are some key points for the correct problem:
- The array is one-dimensional and has been rotated at an unknown pivot.
- You need to determine which half of the array is sorted and then check if the target lies within that sorted half.
- The reference solution provided in C++ can be adapted to Java.

For example, in Java, the solution might look like:

```java
class Solution {
    public int search(int[] nums, int target) {
        int low = 0;
        int high = nums.length - 1;
        
        while (low <= high) {
            if (nums[mid] == target) return mid;
            
            if (nums[low] <= nums[mid]) { // left half is sorted
                if (nums[low] <= target && target < nums[mid]) {
                    high = mid - 1;
                } else {
                    low = mid + 1;
                }
            } else { // right half is sorted
                if (nums[mid] < target && target <= nums[high]) {
                    low = mid + 1;
                } else {
                    high = mid - 1;
                }
            }
        }
        return -1;
    }
}
```

Please ensure you understand the problem requirements before implementing the solution. The problem title and description clearly indicate a rotated sorted array, not a 2D matrix.

sandy7907:
This is search for  a 2D matrix - https://leetcode.com/problems/search-a-2d-matrix/description/. I think you are explaining about search for an element in a rotated sorted array. Both the problem have the same Git project(https://github.com/super30admin/Binary-Search-1).

super30admin:
Hello! It looks like you might have solved a different problem. The problem you solved is "Search a 2D Matrix", where you are given a 2D matrix and need to search for a target. However, the problem you were assigned is "Search inside a Rotated Sorted Array", which involves a one-dimensional array that has been rotated at a pivot.

For the rotated sorted array problem, you need to implement a modified binary search. The key idea is to determine which half of the array is sorted and then check if the target lies within that sorted half. Here's a brief outline of the approach:
- Use binary search with two pointers, `low` and `high`.
- At each step, calculate `mid`.
- Check if the left half (from `low` to `mid`) is sorted by comparing `nums[low]` and `nums[mid]`.
  - If the left half is sorted, check if the target is within the range of the left half. If yes, search left; otherwise, search right.
- Else, the right half (from `mid` to `high`) must be sorted. Check if the target is within the range of the right half. If yes, search right; otherwise, search left.

Your current code is correct for the 2D matrix problem, but please ensure you are solving the correct problem. For the rotated sorted array, you need to work with a 1D array and adjust the binary search logic accordingly.

PR #2470, opened 2026-02-08
Add implementation for Binary-Search-1

rishigoswamy:


super30admin:
Your solution for the rotated sorted array search problem is excellent. You have correctly implemented the binary search by checking which half is sorted and then adjusting the search range based on whether the target lies within the sorted half. The code is efficient and meets the required time and space complexities. The comments are helpful and demonstrate a good understanding of the approach.

One minor suggestion: In the condition `if target >= nums[low] and target < nums[mid]`, you can use parentheses for clarity, although it is not strictly necessary due to operator precedence. For example, `if (target >= nums[low]) and (target < nums[mid])` might be slightly more readable. However, this is a very small point and does not affect the correctness.

Overall, great job!

super30admin:
Overall, the student has a good understanding of binary search and its variations. The solutions for LeetCode 33 and LeetCode 74 are almost correct, with minor issues in boundary conditions. The solution for LeetCode 702 has a more significant logical error in the range expansion.

For LeetCode 33 (leetcode33.py): Well done! This solution is correct and efficient.

For LeetCode 702 (leetcode702.py): The approach is on the right track, but the range expansion loop needs improvement. Instead of using a flag, you can use a while loop that continues until `reader.get(high)` is greater than or equal to the target. Also, you should handle the case where `reader.get(high)` might be out of bounds (but since the array is sorted and has a large size, and the problem states that all values are integers and the array is sorted, you can assume that beyond the array, values are `2^31 - 1` or a large number). Here's a suggested improvement for the range expansion:
```
low = 0
high = 1
    low = high
    high *= 2
    # Also, you might want to check if reader.get(low) is already beyond, but the problem says the array is sorted.
```
Then, once you have the range [low, high], you can call binary search. Also, remember to check for index 0 at the beginning.

For LeetCode 74 (leetcode74.py): The row binary search should be modified to ensure that if no row is found, we return false. Currently, after the row binary search, you break out of the loop when you find a row that might contain the target. But if the target is not in any row, the loop will terminate without breaking, and then `mid` will be the last mid value, which might not be valid. Instead, you should return false if the row binary search ends without finding a candidate row. One way to do this is to check after the loop whether the candidate row actually contains the target in its range. Alternatively, you can structure the row binary search to store the candidate row only when found. For example:
```
low, high = 0, len(matrix)-1
while low <= high:
    mid = (low+high)//2
    if matrix[mid][0] <= target <= matrix[mid][-1]:
        # found row
        break
    elif target < matrix[mid][0]:
        high = mid - 1
    else:
        low = mid + 1
else:
    return False  # no row found
```
Then proceed to search in the row.

In general, the student should pay more attention to edge cases and termination conditions in binary search.

PR #2472, opened 2026-02-10
Problem2 and Problem3 added

megharaykar:


super30admin:
Strengths:
- You have correctly identified the need for binary search to achieve O(log n) time complexity.
- The code structure is clear, and you have handled the basic cases of binary search.
- You have considered the two scenarios (left sorted and right sorted) which is the key to solving this problem.

Areas for Improvement:
1. Function Name: The problem expects a function named `search` to match the LeetCode interface. Rename your function to `search` for compatibility.
2. Condition Checks: The conditions inside the binary search need to be corrected. For the left sorted part, use `if nums[low] <= target and target < nums[mid]` to include the case when target is at the low index. Similarly, for the right sorted part, use `if nums[mid] < target and target <= nums[high]` to include the case when target is at the high index.
3. Edge Cases: Test your solution with edge cases, such as when the target is at the first or last element, or when the array is not rotated. For example, test with `nums = [1,3]` and `target=1` or `target=3`.
4. Remove Extra Code: For LeetCode submissions, avoid including test code (like the instantiation of the class and print statements) in the solution.

Corrected Code Snippet:
```python
class Solution:
    def search(self, nums: List[int], target: int) -> int:
        low, high = 0, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                return mid
            if nums[low] <= nums[mid]:  # left sorted
                if nums[low] <= target < nums[mid]:
                    high = mid - 1
                else:
                    low = mid + 1
            else:  # right sorted
                if nums[mid] < target <= nums[high]:
                    low = mid + 1
                else:
                    high = mid - 1
        return -1
```

super30admin:
Strengths:
- You have understood the core idea of using binary search for the rotated sorted array problem.
- The code structure is clear, and the variables are well-named.
- You correctly handle the base case of finding the target at the mid index.

Areas for Improvement:
- The conditions for checking if the target lies within the sorted half need to be inclusive of the endpoints. For example, in the left sorted part, use `nums[low] <= target < nums[mid]` to include the case when target is exactly at `nums[low]`. Similarly, for the right sorted part, use `nums[mid] < target <= nums[high]`.
- The method name should be `search` instead of `rotatedSorted` to match the problem's interface. This is important for compatibility with automated testing systems.
- Remove the test code at the bottom when submitting the solution.
- The problem requires the solution to be in a class method named `search`. Your solution defines a method `rotatedSorted`, which would not be recognized by the test runner.

Corrected code for Problem2.py should look like:
```python
class Solution:
    def search(self, nums: List[int], target: int) -> int:
        low, high = 0, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                return mid
            if nums[low] <= nums[mid]:
                if nums[low] <= target < nums[mid]:
                    high = mid - 1
                else:
                    low = mid + 1
            else:
                if nums[mid] < target <= nums[high]:
                    low = mid + 1
                else:
                    high = mid - 1
        return -1
```

Note: The above code uses type hints for clarity, which are optional but recommended.

PR #2477, opened 2026-02-27
done binary search 1

sakshiasati17:


super30admin:
Your solutions are well-implemented and efficient. For the first solution (Search2DMatrix), consider using more descriptive variable names: for example, instead of `m_i`, use `mid_index` or `flat_index`, and instead of `t`, you can avoid it by directly using `m*n` in the initialization of `r`. Also, adding a comment explaining the conversion from flat index to 2D indices (i and j) would be helpful. The other two solutions are clear and follow best practices. Keep up the good work!

PR #2478, opened 2026-03-06
Adding the solutions

subbu4061:


super30admin:
The student's solution for "Search in Rotated Sorted Array" is excellent. It correctly implements the binary search algorithm for a rotated sorted array, with appropriate checks to determine which half is sorted and whether the target is within that half. The time and space complexities are optimal. The code is clean and easy to understand.

However, note that the student has included solutions for other problems (SearchA2DMatrix and SearchInInfiniteSortedArray) in the same file. While these are correct for their respective problems, it's important to keep solutions separate for clarity. For the purpose of this evaluation, we are only considering the "Search in Rotated Sorted Array" solution.

One minor point: in the condition `if(nums[low]<= nums[mid])`, it's common to use `<=` to handle the case when `low` and `mid` are the same (which can happen when the array has only one element). This is correctly handled.

Overall, the student has demonstrated a good understanding of binary search and its application to rotated arrays.

PR #2483, opened 2026-04-11
Implemented 2D matrix and search in sorted Array

sumanth00100:


super30admin:
### Search inside a Rotated Sorted Array (search-in-rotated-sorted-array.java)
Your solution is correct and efficient. It meets the problem requirements of O(log n) time and O(1) space. The code is clean and easy to read. Well done!

One minor point: in the condition `if(nums[mid]>= nums[low])`, you are comparing the middle element with the low element to determine if the left part is sorted. This is correct. However, note that when `low == mid`, this condition holds true (since it's equal), which is correct because a single element is sorted. So your condition is robust.

You might consider adding comments within the code to explain the two cases (left sorted and right sorted) to make it more understandable for others. For example:

```
if (nums[mid] >= nums[low]) { // left part is sorted
    if (target >= nums[low] && target < nums[mid]) {
        // target is in the left sorted part
        high = mid - 1;
    } else {
        // target is not in the left sorted part, so search right
        low = mid + 1;
    }
} else { // right part is sorted
    if (target <= nums[high] && target > nums[mid]) {
        // target is in the right sorted part
        low = mid + 1;
    } else {
        // target is not in the right sorted part, so search left
        high = mid - 1;
    }
}
```

This would enhance readability.

Overall, great job!

VERDICT: PASS

---

### Search Inside a Sorted Array whose Length is unknown (2dMatrix.java)
It seems there might be a confusion in the problem you are solving. The problem you are addressing is "Search a 2D Matrix", but the problem you were supposed to solve is "Search in a Sorted Array of Unknown Size". The latter requires using an ArrayReader interface, which is not present in your code. 

1. Use the ArrayReader interface to access elements.
2. First, find a range where the target might exist by doubling the index until you get a value greater than or equal to the target (or hit the boundary).
3. Then, perform binary search within that range.

Your current solution is correct for the 2D matrix problem, but it does not solve the problem intended here. Please review the problem statement again and implement the solution accordingly.

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix
Your solution is well-implemented and efficient. You correctly applied binary search by converting the 2D indices into 1D indices. Here are a few suggestions for improvement:
1. Consider renaming "matrixLen" to "cols" or "n" to make it clearer that it represents the number of columns.
2. Although not required by the constraints, it's good practice to add a check for an empty matrix (if matrix.length == 0) at the beginning to make the code more robust.

Overall, great job! Your code is concise and meets all the problem requirements.

VERDICT: PASS

PR #2487, opened 2026-04-13
Binary-Search-1 Solution

Simranb10:


super30admin:
### Search inside a Rotated Sorted Array (problem1Solution.java)
Your solution is well-written for problem 74 (Search a 2D Matrix), but it does not solve the problem you were given. The problem is about searching in a rotated sorted array, which requires a different approach.

1. Recognize that the array is rotated, meaning there are two sorted portions.
2. Use a modified binary search to determine which side is sorted and then check if the target is within that sorted side.
3. The method signature should be: `public int search(int[] nums, int target)`

I suggest you revisit the problem statement and try to implement the solution for the rotated sorted array. The reference solution provided can guide you.

VERDICT: NEEDS_IMPROVEMENT

---

### Search Inside a Sorted Array whose Length is unknown (problem2Solution.java)
Your solution is for a different problem (Search in a Rotated Sorted Array) and does not meet the requirements of this problem. Here are the key issues:
1. The problem requires interacting with an ArrayReader interface, but your solution uses a fixed-length array (int[] nums) as input. You should instead use the ArrayReader methods to access elements.
2. The problem states that the array length is unknown, so you cannot use nums.length. You need to first find a range where the target might lie by exponentially expanding the search range until you find an upper bound that is beyond the target.
3. Your current solution assumes the array is rotated, but the problem states the array is strictly increasing (sorted), so rotation is not a factor.

To correct this:
- You should start by setting low=0 and high=1, then while ArrayReader.get(high) is less than the target, set low=high and double high. This finds a range that contains the target.
- Then perform binary search between low and high.
- Remember to handle the case when ArrayReader.get(i) returns 2^31-1 for out-of-bound indices.

Here is a corrected version in Java for your reference:
```
public class Solution extends ArrayReader {
    public int search(ArrayReader reader, int target) {
        int low = 0, high = 1;
        while (reader.get(high) < target) {
            low = high;
            high *= 2;
        }
        while (low <= high) {
            int midVal = reader.get(mid);
            if (midVal == target) return mid;
            if (midVal > target) {
                high = mid - 1;
            } else {
                low = mid + 1;
            }
        }
        return -1;
    }
}
```

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix (problem3Solution.java)
Your solution is for a different problem (Search in Infinite sorted array) and does not solve the 2D matrix search problem. Here are the main issues:
1. The problem requires an input of a 2D integer matrix and a target integer, but your method signature is `search(ArrayReader reader, int target)`, which is incorrect.
2. You are using an `ArrayReader` class which is not defined in the problem context.
3. The approach you implemented is for an infinite sorted array, which is not applicable here. The 2D matrix has fixed dimensions and specific properties that allow for a binary search over the entire matrix as if it were a 1D sorted array.

To correct this:
- You should use the correct method signature: `public boolean searchMatrix(int[][] matrix, int target)`.
- The solution should first check the dimensions of the matrix (m rows and n columns).
- Then, perform a binary search over the range [0, m*n-1], treating the matrix as a 1D array. For each mid index, compute the row and column indices using `row = mid / n` and `col = mid % n`.
- Compare the element at matrix[row][col] with the target and adjust the search range accordingly.

Your current code has good structure for binary search, but it is applied to the wrong problem. Please review the problem statement and the reference solution provided.

VERDICT: NEEDS_IMPROVEMENT

Simranb10:
I followed the order given in the README file. I’ll move the code to the correct files and push the changes.

PR #2489, opened 2026-04-14
Solved Binary Search Set 1

pratikb0501:


super30admin:
### Search inside a Rotated Sorted Array (Problem_1.py)
Your solution for searching in a 2D matrix is well-written and efficient for that problem. However, the problem you were asked to solve is about searching in a rotated sorted array, not a 2D matrix. It seems there might have been a misunderstanding of the problem statement.

For the rotated sorted array problem, you need to implement a binary search that accounts for the rotation. The key idea is to determine which half of the array is sorted and then check if the target lies within that sorted half. Here's a brief outline of the correct approach:
1. Initialize `low = 0` and `high = len(nums)-1`.
2. While `low <= high`:
   - Calculate `mid = low + (high - low) // 2`.
   - If `nums[mid] == target`, return `mid`.
   - Check if the left half (from `low` to `mid`) is sorted by comparing `nums[low]` and `nums[mid]`.
   - If the left half is sorted, check if the target is within this range. If yes, search left; otherwise, search right.
   - If the right half is sorted, check if the target is within the right half. If yes, search right; otherwise, search left.

Your current code does not address this logic at all. I recommend revisiting the problem statement and implementing the solution for the rotated sorted array. Also, ensure that your function name and parameters match the requirements (e.g., the function should be named `search` and take `nums` and `target` as parameters).

VERDICT: NEEDS_IMPROVEMENT

---

### Search Inside a Sorted Array whose Length is unknown (Problem_2.py)
Your solution demonstrates a good understanding of binary search for rotated sorted arrays, but it does not address the problem given. The problem requires you to work with an `ArrayReader` interface because the array length is unknown. You need to first find the range where the target might exist by exponentially expanding the search window until you find an upper bound that is beyond the target. Then, perform a standard binary search within that range.

Here are the key issues:
1. The problem specifies that the array length is unknown and access is through `ArrayReader.get(i)`, but your solution uses a list `nums` with known length.
2. Your solution includes a test case at the end, which should not be part of the solution class. The code should only contain the `search` method.

To correct this:
- The method signature should be `def search(self, reader: ArrayReader, target: int) -> int:`.
- You need to define a way to find the high bound. Start with low=0 and high=1, and while `reader.get(high)` is less than the target, set low=high and double high. Then, perform binary search between low and high.
- Remember to handle the case when `reader.get(i)` returns `2^31 - 1` (which is 2147483647) for out-of-bound indices.

Example corrected approach:
```python
class Solution:
    def search(self, reader, target):
        low, high = 0, 1
            low = high
            high *= 2
        while low <= high:
            mid = low + (high - low) // 2
            val = reader.get(mid)
            if val == target:
                return mid
            elif val > target:
                high = mid - 1
            else:
                low = mid + 1
        return -1
```

Please revise your solution to match the problem requirements.

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix (Problem_3.py)
Note: The verdict should be PASS if the solution is correct and efficient, otherwise NEEDS_IMPROVEMENT.

Note: The student's solution is for a different problem (Search in a Sorted Array of Unknown Size) and not for the Search a 2D Matrix problem. The student's code is for a problem where the array size is unknown and uses an ArrayReader interface, which is not relevant to the given problem.

Therefore, the student's solution does not address the problem at all.

Let's evaluate accordingly.

VERDICT: NEEDS_IMPROVEMENT

PR #2490, opened 2026-04-23
Binary search 1 solutions

indrasena483:


super30admin:
### Search inside a Rotated Sorted Array (SearchUnknownSizeArray.java)
Your solution is well-written for the "Search in a Sorted Array of Unknown Size" problem. However, this is not the problem that was assigned. The problem you should solve is "Search in Rotated Sorted Array", where the array has been rotated at an unknown pivot.

For the correct problem, you need to account for the rotation. The key idea is to use binary search to determine which half of the array is sorted (left or right) and then check if the target lies within that sorted half. Here's a breakdown of the approach:
1. Initialize low = 0 and high = nums.size() - 1.
2. While low <= high:
   a. Calculate mid = low + (high - low) / 2.
   b. If nums[mid] equals target, return mid.
   c. Check if the left half [low, mid] is sorted (i.e., nums[low] <= nums[mid]).
      - If yes, then check if target is in [nums[low], nums[mid]): if so, search left; else search right.
   d. Otherwise, the right half [mid, high] must be sorted. Then check if target is in (nums[mid], nums[high]]: if so, search right; else search left.

You must implement this logic because the array is rotated. Your current solution does not handle rotation.

Please revise your code to address the correct problem. You can use the reference solution provided as a guide.

VERDICT: NEEDS_IMPROVEMENT

---

### Search Inside a Sorted Array whose Length is unknown (SearhRotatedArray.java)
It seems there might be a misunderstanding. The problem you solved (#33. Search in Rotated Sorted Array) is different from the one you were asked to solve (Search Inside a Sorted Array whose Length is unknown). The given problem requires you to work with an ArrayReader interface because the array length is unknown. You need to first find the bounds of the array and then perform binary search within those bounds.

For the correct problem, you should:
1. Start with a low index (0) and a high index (1).
2. Expand the high index exponentially (e.g., high = high * 2) until you get a value that is greater than or equal to the target (or until you get the out-of-boundary value).
3. Then perform binary search between low and high.

Your current code does not use the ArrayReader interface and assumes the array length is known. Please revisit the problem statement and implement the solution accordingly.

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix (Search2DMatrix.java)
**

Great job! Your solution is correct and efficient. You've successfully implemented the binary search algorithm for a 2D matrix, achieving the required O(log(m*n)) time complexity and O(1) space complexity. Your code is clean and well-commented.

Here are a few minor suggestions for improvement:
1. **Code Formatting:** Pay attention to consistent spacing and brace placement. For example, use a space after `while` and before the opening parenthesis: `while (low <= high) {`. Also, consider placing the closing brace on a new line for better readability.
2. **Comments:** Your header comments are good, but you could add a brief description of the algorithm (e.g., "This solution treats the 2D matrix as a 1D array and performs binary search.") to make it even clearer.

Overall, your solution is excellent and meets all the requirements.

**

VERDICT: NEEDS_IMPROVEMENT

PR #2492, opened 2026-05-03
Solved Problem 1 - Seach a 2d Matrix, Problem 2 - Search in a rotated sorted array and Problem 3 - Search in infinite sorted array

anirudhv98:
Completed Binary-Search-1 - Please review


super30admin:
### Search inside a Rotated Sorted Array (Problem1.cs)
It seems there might be a misunderstanding. The problem you solved is "Search a 2D Matrix", but the problem you were asked to solve is "Search inside a Rotated Sorted Array". These are two distinct problems. 

For the rotated sorted array problem, you need to implement a binary search that accounts for the rotation. The key idea is to determine which half of the array is sorted and then check if the target lies within that sorted half. If it does, you continue the search in that half; otherwise, you search the other half.

Here are the steps you should follow for the correct problem:
1. Initialize `low` to 0 and `high` to `nums.length - 1`.
2. While `low <= high`, calculate `mid`.
3. If `nums[mid]` equals the target, return `mid`.
4. Check if the left half (from `low` to `mid`) is sorted (i.e., `nums[low] <= nums[mid]`).
   - If it is sorted, check if the target is within the range `[nums[low], nums[mid])`.
     - If yes, then set `high = mid - 1` to search the left half.
     - Else, set `low = mid + 1` to search the right half.
5. Else, the right half (from `mid` to `high`) must be sorted.
   - Check if the target is within the range `(nums[mid], nums[high]]`.
     - If yes, then set `low = mid + 1` to search the right half.
     - Else, set `high = mid - 1` to search the left half.

Your current solution does not address this logic. Please revisit the problem statement and implement the correct algorithm for the rotated sorted array.

VERDICT: NEEDS_IMPROVEMENT

---

### Search Inside a Sorted Array whose Length is unknown (Problem2.cs)
Your solution is well-written for the problem of searching in a rotated sorted array. However, it does not solve the problem presented here. The current problem requires handling an array of unknown size using the ArrayReader interface, which your code does not utilize. Instead, your code assumes a known array length and checks for rotation, which is not part of this problem.

To correct this, you need to:
1. Use the ArrayReader interface instead of an array parameter. The method signature should be something like `Search(ArrayReader reader, int target)`.
2. First, find a range where the target might lie by exponentially increasing the index until you find an index where the value is greater than or equal to the target (or out of bounds).
3. Then, perform binary search within that range.

For example, you can start with low=0 and high=1, and while `reader.get(high) < target`, set low=high and high=high*2. Then, once you have a range [low, high] where the target might be, perform binary search.

Please adjust your solution to address the correct problem. Your current approach is correct for a rotated sorted array, but not for this problem.

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix (Problem3.cs)
It seems there might be a mix-up in the problems. Your solution is for "Search in a Sorted Array of Unknown Size", but the problem you were asked to solve is "Search a 2D Matrix". 

For the "Search a 2D Matrix" problem, you need to implement a function that takes a 2D integer matrix and an integer target. The matrix has the properties that each row is sorted in non-decreasing order, and the first integer of each row is greater than the last integer of the previous row. This allows us to treat the entire matrix as a single sorted array and perform binary search over it.

Here's how you can adapt your approach:
1. Instead of ArrayReader, you have a 2D array `matrix`.
2. The total number of elements is `m * n`, where `m` is the number of rows and `n` is the number of columns.
3. You can set `low = 0` and `high = m * n - 1`.
4. For a given index `mid` in the flattened array, you can compute the row index by `mid / n` and the column index by `mid % n`.
5. Then compare `matrix[r][c]` with the target.

Your current solution for the unknown size array is good, but please ensure you are solving the correct problem. For the 2D matrix problem, you should write a solution similar to the reference provided.

VERDICT: NEEDS_IMPROVEMENT

PR #2497, opened 2026-05-24
Binary Search - 1

KartavyaBhatt:


super30admin:
### Search inside a Rotated Sorted Array (Problem_1.py)
1. **Wrong Problem:** You implemented a solution for "Search a 2D Matrix" instead of "Search in Rotated Sorted Array". The problem asks you to find a target in a single rotated sorted array, not a 2D matrix.

2. **Understanding the Problem:** Make sure to carefully read the problem description. The input is a single array `nums` that may be rotated, not a 2D matrix.

3. **Key Insight for Rotated Array:** The solution should use binary search where you determine which half of the array is properly sorted at each step, then decide which half to search based on where the target lies.

4. **Strengths:** Your code is well-organized with clear method separation and follows good practices (avoiding integer overflow with `l + (r-l)//2`). Your comments are helpful.

5. **Next Steps:** Study the rotated sorted array problem - the key is to check which half of the array is sorted and whether the target falls within the sorted half's range.

VERDICT: NEEDS_IMPROVEMENT

---

### Search Inside a Sorted Array whose Length is unknown (Problem_2.py)
**Strengths:**
- Good understanding of binary search concepts
- Clean recursive implementation with proper base case
- Good variable naming and code structure
- Proper use of `l + (r-l)//2` to avoid integer overflow

**Areas for Improvement:**
1. **Wrong Problem**: The solution solves "Search in Rotated Sorted Array" instead of the intended problem about searching in an unknown-length array using `ArrayReader`
2. **Missing Interface**: The solution doesn't use the `ArrayReader` interface at all - it directly accesses array elements
3. **Out of Bounds Handling**: The original problem requires handling the case where `reader.get(i)` returns `2^31 - 1` for out-of-bounds indices
4. **Space Complexity**: The recursive approach uses O(log n) space due to call stack; could be optimized to O(1) with iteration
5. **Unknown Length Handling**: The original problem requires finding the array bounds first before doing binary search

To fix this, you need to:
1. Use `ArrayReader.get(i)` instead of direct array access
2. Implement exponential search to find the bounds first
3. Then perform binary search within those bounds
4. Consider using an iterative approach for O(1) space complexity

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix (Problem_3.py)
**

The student's solution demonstrates strong understanding of binary search and exponential search techniques. The code is clean and well-commented. However, the solution addresses a different problem entirely. The student needs to:

1. Understand the exact problem requirements - this problem involves a 2D matrix, not an ArrayReader
2. Implement the 1D-to-2D index mapping using `mid / n` for row and `mid % n` for column
3. Apply binary search directly on the flattened matrix representation

**

VERDICT: NEEDS_IMPROVEMENT

PR #2509, opened 2026-06-08
Rotational sort and unknown range

ADITYABHURAN:


super30admin:
### Search inside a Rotated Sorted Array (problem2.py)
**Strengths:**
- Clean and well-structured implementation of binary search
- Excellent documentation with time/space complexity comments
- Correct handling of both sorted halves in the rotated array
- Pythonic code with proper integer division syntax
- The solution is functionally equivalent to the reference solution

**Areas for Improvement:**
- The comment "#left sorted" appears twice (lines 11 and 13) - could be more descriptive
- Consider adding a brief docstring explaining the algorithm approach
- The code could benefit from more detailed inline comments explaining the binary search logic for rotated arrays

The solution is solid and correctly solves the problem. It matches the reference solution in both approach and efficiency.

VERDICT: PASS

---

### Search Inside a Sorted Array whose Length is unknown (problem3.py)
**Strengths:**
- Good understanding of the overall approach (exponential search to find range + binary search)
- Clean variable naming (low, high, mid)
- Comments explaining the approach

**Areas for Improvement:**
- Fix the indentation issue - the binary search logic must be inside the while loop
- Simplify the binary search - since the array is strictly sorted, there's no need for the additional orientation checks
- Remove the redundant `reader.get(low)` calls
- The comment about time complexity is incorrect; it should be O(log n) not O(log m + log n)

**Corrected approach:**
Since the array is guaranteed to be sorted in strictly increasing order, the binary search portion can be simplified to just:
```python
while low <= high:
    mid = low + (high - low) // 2
    if reader.get(mid) == target:
        return mid
    elif reader.get(mid) > target:
        high = mid - 1
    else:
        low = mid + 1
return -1
```

VERDICT: NEEDS_IMPROVEMENT

---

### Search a 2D Matrix
The student appears to have submitted solutions for different LeetCode problems rather than the "Search a 2D Matrix" problem. The reference solution demonstrates the correct approach: treating the 2D matrix as a 1D sorted array and using binary search with proper index mapping (row = mid // n, col = mid % n). 

Key issues:
- Wrong problem solved (rotated array and infinite array instead of 2D matrix search)
- `problem3.py` has indentation errors in the binary search section
- The solution doesn't leverage the matrix property that it can be viewed as a single sorted array
- Time complexity analysis doesn't match the problem requirements

The student should focus on understanding how to map a 1D binary search index to 2D matrix coordinates.

VERDICT: NEEDS_IMPROVEMENT

